PR #39, opened 2020-05-15
28 divider component

vlineros:
Added divider component and storybook knobs to control height and width
Also added a themed example


packages/hs-react-ui/src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import styled, { StyledComponentBase } from 'styled-components';
+
+import colors from '../../constants/colors';
+
+export const DefaultDivider = styled.hr`
+  ${({ width = 90, height = 1}: {width: number, height: number}) => `
+    border: none;
+    height: ${height}px;
+    width: ${width}%;
```
aVileBroker, line 10: let's let them pass in a string so they can choose whatever units they want (we can add the units to the knob number in the story)

PR #41, opened 2020-05-15
Checkbox Component

sambskn:
Wasn't able to get to the Radio inputs yet, but have the Checkbox component working.


aVileBroker:
Still reviewing - looks like you used npm instead of yarn, should be easy to avoid by either just deleting your package-lock.json or git ignoring it so it isn't a problem for anyone in the future :)

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const Checkbox = ({
+    checkboxType,
+    checked,
+    children,
+}: CheckboxProps) => {
+    const iconPath = iconPathFromType(checkboxType)
+    const IconComponent = iconComponentFromType(checkboxType)
+    return(
+        <div>
```
aVileBroker, line 126: what's this div for? o.O

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,140 @@
...
+            return CrossIcon;
+        case CheckboxTypes.neutral:
+            return NeutralIcon;    
+        default:
+            return DefaultIcon;
+    }
+}
+
+const Checkbox = ({
+    checkboxType,
```
aVileBroker, line 119: If you default checkboxType to `default`, you can just access the icon from an object, like `checkboxPaths[checkboxType]`

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,140 @@
...
+    checked,
+    children,
+}: CheckboxProps) => {
+    const iconPath = iconPathFromType(checkboxType)
+    const IconComponent = iconComponentFromType(checkboxType)
+    return(
+        <div>
+            <Label>
```
aVileBroker, line 127: If there's no `children` being passed in, is the `Label` wrapper needed? Just trying to slim down our DOM/style footprint

aVileBroker: But then again we don't want it to look significantly different just because they passed in children

packages/hs-react-ui/src/enums/CheckboxTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+    cross = 'cross',
+    check = 'check',
+    default = 'default',
+    neutral = 'neutral'
+}
+
+  
+// export enum RadioTypes {
+//     default = 'default'
+// }
```
aVileBroker, line 12: use it or lose it imo

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    ${Input}:focus + & {
+        box-shadow: 0 0 0 3px ${colors.grayXlight};
+    }
+`;
+
+const Box = styled.div`
+    border: 1px solid  ${colors.blueCharcoalXLight};
+    border-radius: 2px;
+    // background-color: ${colors.background};
```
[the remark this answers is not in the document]
sambskn: u right

packages/hs-react-ui/src/constants/colors.ts
```diff
@@ -1,11 +1,14 @@
...
   primary: '#e75b27',
-  
+
   grayDark: '#252c35',
   grayMedium: '#51575d',
   grayLight: '#7c8186',
   grayXlight: '#d3d6d7',
 
+  blueCharcoalLight: '#51575D',
+  blueCharcoalXLight: '#7C8186',
```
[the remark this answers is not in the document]
sambskn: oh did not check the codes for comparisons, my b. you should have called them blueCharcoal* then lol

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
sambskn: Label wrapper is needed when we're overriding the default input so we can still use an onChange handler effectively. I wanted to let them pass in children so they could have other elements (maybe an icon or whatever) that would still trigger an onChange for the checkbox when clicked because they are wrapped by the label.

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,140 @@
...
+    height: 0.7rem;
+    width: 0.7rem;
+`;
+
+const NeutralIcon = styled(StyledIcon)`
+    color: ${colors.blueCharcoalLight};
+    height: 0.65rem;
+    width: 0.65rem;
+    path {
+        stroke: ${colors.blueCharcoalLight};
```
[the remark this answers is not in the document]
sambskn: ok

packages/hs-react-ui/src/enums/CheckboxTypes.ts
sambskn: ok

packages/hs-react-ui/src/components/Checkbox/Checkbox.tsx
sambskn: good point lol

sambskn:
Updated with fixes suggested by @aVileBroker

PR #50, opened 2020-05-26
Add jest and coverage reporting

joien1:


.circleci/config.yml
```diff
@@ -11,6 +11,7 @@ jobs:
       - node/with-cache:
           steps:
             - run: npm run build-react
+            - run: npm run test
```
joien1, line 14: This is just a wild guess at how this might work, @aVileBroker

aVileBroker: So the reason it's build-react is that inside that script it changes directory into the hs-react-ui folder and then runs a bunch of scripts. So either add a test-react script which does the same thing, or add to the build-react script.

PR #53, opened 2020-05-29
27 label enhance

vlineros:
Label now has input to be 'for' as well as required, valid and invalid states.
states include icon and color adjustment
added a boolean to control checking for validity


packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
 import colors from '../../enums/colors';
 
 export const StyledLabel = styled.label`
-  ${({ color = colors.grayDark }: { color: colors | string }) => `
-    display: block;
+  ${({ color = colors.grayLight }: { color: colors | string }) => `
+    display: inline-flex;
     color: ${color};
-    text-transform: uppercase;
+    font-family: Arial, Helvetica, sans-serif;
```
aVileBroker, line 11: If you look at some of the other files, you'll see the `{fonts.body}` style - we should use that going forward to get the correct fonts.

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+export const Child = styled.input`
+  ${({ color = colors.grayLight }: { color: colors | string }) => `
+  display: block;  
+  border: solid 2px ${color};
+  width: 15%;
   `}
 `;
 
+export const StyledDiv = styled.div`
```
aVileBroker, line 25: This should be called LabelContainer to make the API a bit more descriptive

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+      </StyledLabel>
+      <IconContainer>
+      <Icon path={shownIcon} size="11px" color={shownColor}/>
+      </IconContainer>
+      <Child id={htmlFor} color={shownColor}/>
+    </StyledDiv>
+    )
+};
 
 export default Label;
```
aVileBroker, line 83: Check some of the other files - like `Card` - we should add the subcomponents to the main Component variable before exporting it so they're all available when importing the Label

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+  `}
+`;
+
+export const Child = styled.input`
+  ${({ color = colors.grayLight }: { color: colors | string }) => `
+  display: block;  
+  border: solid 2px ${color};
+  width: 15%;
   `}
 `;
```
aVileBroker, line 23: You should let them import whatever child they want (use the `children` prop) instead of making an input here.

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+    return(
+    <StyledDiv>
+      <StyledLabel htmlFor={htmlFor} color={shownColor}>
+        {labelText}
+      </StyledLabel>
+      <IconContainer>
+      <Icon path={shownIcon} size="11px" color={shownColor}/>
+      </IconContainer>
+      <Child id={htmlFor} color={shownColor}/>
```
aVileBroker, line 78: This is where you'd pass the provided children through for rendering

This will mean that the validation won't be able to be passed down, but it does allow the end dev to add validation styling before they pass it into the Label.
```suggestion
      {children}
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+const Label = ({
+  labelText,
+  color = colors.grayLight,
+  isValid = false,
+  checkValidity = false,
+  colorValid = colors.success,
+  colorInvalid = colors.destructive,
+  htmlFor = 'default',
+  isRequired = false,
+}: LabelProps) => {
```
aVileBroker, line 54: add a children prop so the end dev can pass whatever they want inside.
```suggestion
const Label = ({
  labelText,
  color = colors.grayLight,
  isValid = false,
  checkValidity = false,
  colorValid = colors.success,
  colorInvalid = colors.destructive,
  htmlFor = 'default',
  isRequired = false,
  children
}: LabelProps) => {
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
 }
 
-const Label = ({ labelText, color = colors.grayDark }: LabelProps) => (
-  <StyledLabel color={color}>{labelText}</StyledLabel>
-);
+const Label = ({
+  labelText,
+  color = colors.grayLight,
+  isValid = false,
+  checkValidity = false,
```
aVileBroker, line 49: This may not be needed if you don't default isValid to false. That way you can just display validation if `isValid !== undefined`
```suggestion
  isValid,
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+const IconContainer = styled.span`
+  display: inline-flex;
+  margin-left: 0.25rem;
+`;
+
 export interface LabelProps {
   labelText: string;
   color?: colors | string;
+  isValid: boolean;
+  checkValidity: boolean;
```
aVileBroker, line 38: This can be updated so that it accepts undefined
```suggestion
  isValid?: boolean;
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+  isRequired = false,
+}: LabelProps) => {
+    let shownColor: string | colors;
+    let shownIcon: string | JSX.Element;
+
+    if (checkValidity) {
+      shownColor = isValid ? colorValid : colorInvalid;
+    } else {
+      shownColor = color;
+    }
```
aVileBroker, line 62: We can now simplify this to
```suggestion
    if (isValid === true) {
      shownColor = colorValid;
      shownIcon = mdiCheckBold;
    } else if (isValid === false) {
      shownColor = colorInvalid;
    } else {
      shownColor = color;
      shownIcon = isRequired ? mdiAsterisk : null;
    }
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+    if (isRequired) {
+      shownIcon = checkValidity && isValid ? mdiCheckBold : mdiAsterisk;
+    } else {
+      shownIcon = checkValidity && isValid ? mdiCheckBold : '';
+    }
```
aVileBroker, line 68: We can move this into the previous set of if-statements
```suggestion
```

packages/hs-react-ui/src/components/Label/Label.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+    return(
+    <StyledDiv>
+      <StyledLabel htmlFor={htmlFor} color={shownColor}>
+        {labelText}
+      </StyledLabel>
+      <IconContainer>
+      <Icon path={shownIcon} size="11px" color={shownColor}/>
```
aVileBroker, line 76: 11px is probably closest to 12px, which we can represent with a `.75rem` - remember we want to round everything to a whole/half/quarter rem to stick to our rem grid.

This also could use an extra level of indention.
```suggestion
        <Icon path={shownIcon} size=".75rem" color={shownColor}/>
```

packages/hs-react-ui/src/components/Label/Label.stories.tsx
```diff
@@ -20,7 +20,13 @@ storiesOf('Label', module).add(
...
     <Label
       labelText={text('labelText', 'This is the label text')}
-      color={select('Color', colors, colors.grayDark)}
+      color={select('Color', colors, colors.grayLight)}
+      colorValid={select('Color-valid', colors, colors.success)}
+      colorInvalid={select('Color-invalid', colors, colors.destructive)}
+      isRequired={boolean('isRequired', false)}
+      checkValidity={boolean('checkValidity', false)}
+      isValid={boolean('isValid', false)}
+      htmlFor="default"
```
aVileBroker, line 29: let's make this a text knob, just so they know they can change it from this page.
```suggestion
      htmlFor={text('htmlFor', 'default')}
```

packages/hs-react-ui/src/components/Label/Label.stories.tsx
```diff
@@ -20,7 +20,13 @@ storiesOf('Label', module).add(
...
     />
```
aVileBroker, line 30: now that we're accepting the children, open up component to have an hs-ui textInput inside. Something like this, but probably more detail:

```suggestion
    >
      <TextInput id={text('htmlFor', 'default')} />
    </Label>
```

vlineros: To get this working I had to add the id prop to textInput

packages/hs-react-ui/src/components/Label/Label.tsx
vlineros: Had to bring in master to get the fonts.body enum

vlineros:
@aVileBroker Currently have all the changes in from the PR review. Haven't tackled testing yet but I know Marco is looking to use the label for what he is doing so was going to ask if we could merge this now and I can get the testing in on a different PR. Let me know!

PR #57, opened 2020-06-01
HSUI-32 Table enhancements

donaldjbrady:
* Added support for grouped rows
* Added support for group "labels" that can be positioned above or below the group
* Added support to sort Tables with grouped content in two ways. Grouped tables are, by default, sorted by only sorting the content of each group. Optionally, use the `sortGroups` property to sort the contents of the groups as well as the groups. Groups are ordered based on the first element contained in the sorted column for that group.
* Ability to make groups collapsable with the property `isCollapsable`
* Support custom icons for the icon used to collapse and expand the Table groups


packages/hs-react-ui/src/components/Table/Table.stories.tsx
```diff
@@ -35,7 +50,18 @@ const NoteField = styled.textarea`
   resize: none;
 `;
 
-const generateSampleData = (rows: number) => {
+const groupLabelPositionOptions: { above: 'above'; below: 'below' } = {
+  above: 'above',
+  below: 'below',
+};
```
aVileBroker, line 56: Does this need to be typed?

packages/hs-react-ui/src/components/Table/Table.stories.tsx
```diff
@@ -182,3 +223,136 @@ storiesOf('Table', module).add(
...
+  'Groups',
+  () => {
+    const [rows, setRows] = useState(sampleGroupData);
+
+    const onSelect = (index: number, groupIndex = 0, selected: boolean) => {
+      const newRows: Array<SampleSelectionCellType[]> = [];
+      rows.forEach((grp: SampleSelectionCellType[]) => {
+        newRows.push([...grp]);
+      });
+      // const newRows = [...rows];
```
aVileBroker, line 237: I assume we can delete this?

packages/hs-react-ui/src/components/Table/Table.stories.tsx
```diff
@@ -182,3 +223,136 @@ storiesOf('Table', module).add(
...
+    );
+
+    const sampleColumns: { [index: string]: any } = {
+      selection: {
+        name: '',
+        headerCellComponent: SelectAllCell,
+        width: text('Selection width', '2rem'),
+        cellComponent: SelectionCell,
+        sortable: false,
+        groupCellComponent: EmptyCell,
```
aVileBroker, line 320: Is an empty cell required for the groupCellComponent if a cellComponent is defined?

packages/hs-react-ui/src/components/Table/Table.stories.tsx
```diff
@@ -182,3 +223,136 @@ storiesOf('Table', module).add(
...
+    };
+    const position = select('groupLabelPosition', groupLabelPositionOptions, 'above');
+    return (
+      <Table
+        columns={sampleColumns}
+        data={rows}
+        sortGroups={boolean('sortGroups', false)}
+        groupHeaderPosition={position}
+        isCollapsable={boolean('isCollapsable', false)}
+        minWidthBreakpoint={0}
```
aVileBroker, line 351: Do groups not work with the responsive version of the table? We should probably default to 0 or -1 or something so we don't have to pass this just to avoid the mobile version.

but groups should probably support responsive modes as well

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -1,48 +1,83 @@
...
-import { mdiArrowDown } from '@mdi/js';
-
 import colors from '../../enums/colors';
 import fonts from '../../enums/fonts';
 
+/** Types and Interfaces */
+
+export type CellOptions = {
+  RenderedCell: any;
+  headerColumnKey: string; //
```
aVileBroker, line 13: this looks like it was going to be a comment at some point

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -1,48 +1,83 @@
...
+  breakPointHit: boolean;
+  row: columnTypes;
+  index: number;
+  indexModifier?: number;
+  CollapseIcon?: any;
+  groupIndex?: number;
+  isCollapsed?: boolean;
+  cIcon?: any;
+  eIcon?: any;
```
aVileBroker, line 22: what are these for? Collapsed/Expanded icons? If so, the prevoius CollapseIcon prop is probably a duplicate? Either way we should use full prop names like CollapseIcon to make them easy to understand

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -1,48 +1,83 @@
...
+    isGroupLabel?: boolean;
     cellComponent?: any;
     rowComponent?: any;
     headerCellComponent?: any;
+    groupCellComponent?: any;
   };
 }
 
 export type TableProps = {
+  isCollapsable?: boolean;
```
aVileBroker, line 50: `areGroupsCollapsable` might make more sense, the table itself doesn't collapse

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -1,48 +1,83 @@
...
+  groupIndex?: number;
+  isCollapsed?: boolean;
+  cIcon?: any;
+  eIcon?: any;
+  groupLabelDataString?: string;
+};
+
+export type ExpansionIconProps = {
+  collapsedIcon: string;
+  expandedIcon: string;
```
aVileBroker, line 28: should these be `any`?

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -1,48 +1,83 @@
...
-  defaultSort?: [string, boolean]; // key, direction
-  data?: columnTypes[];
   columns: columnTypes;
-
+  data?: columnTypes[] | Array<Array<columnTypes>>;
+  defaultSort?: [string, boolean]; // key, direction
+  groupHeaderPosition?: 'above' | 'below';
+  expansionIconComponent?: React.FunctionComponent<ExpansionIconProps>;
+  collapsedIcon?: string | React.FunctionComponent<any>;
+  expandedIcon?: string | React.FunctionComponent<any>;
```
aVileBroker, line 58: Should we just allow anything to be passed in as an icon and then pass it the data it needs to render (isCollapsed)

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -132,6 +167,10 @@ export const Row = styled.tr`
   `}
 `;
 
+export const GroupRow = styled(Row)`
+  background-color: #e8e8e8;
```
aVileBroker, line 171: color constant

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -151,31 +190,74 @@ const SortIcon = styled(Icon)`
...
+const ExpansionIcon: React.FunctionComponent<ExpansionIconProps> = ({
+  collapsedIcon,
+  expandedIcon,
+  isCollapsed,
+  onClick,
+}: ExpansionIconProps) => {
+  const path = isCollapsed ? collapsedIcon : expandedIcon;
+  return (
+    <span tabIndex={0} onClick={onClick} role="button" onKeyPress={onKeyPress}>
+      <Icon path={path} size="1rem" />
```
aVileBroker, line 220: Let's default to Icon but we should let them pass any component for collapsed/expanded - we don't want to force mdi on the end developer

packages/hs-react-ui/src/components/Table/Table.stories.tsx
donaldjbrady: They need to in order to pass the linting on the build. The inputs for the label position are typed as one of the strings `above` or `below` and without the typing they get flagged as type string rather than `above` or `below`

packages/hs-react-ui/src/components/Table/Table.stories.tsx
donaldjbrady: The intention is if `cellComponent` is defined and `groupCellComponent` is not, then `cellComponent` is used. `GroupCellComponent` should be used to override the `CellComponent` for the group 'header.' I meant for it to work like this, but just checked the implementation and that needs corrected.

packages/hs-react-ui/src/components/Table/Table.stories.tsx
donaldjbrady: You know what, I completely forgot about this after I got everything working. I was having issues with the groups when all of the groups were collapsed. I'll check more into this

packages/hs-react-ui/src/components/Table/Table.tsx
donaldjbrady: These are the overrides for the expanded and collapsed icons. It's how a developer can pass material Icons (or any svg path) of their choice into the Table without providing their own custom expansion Icon too.

packages/hs-react-ui/src/components/Table/Table.tsx
donaldjbrady: These are for the internal `ExpansionIcon` component that shows the arrow down/right by default. The developer has the ability to either pass in a string path to override the default without anything other than passing in a string of the path for a material icon - or if they choose they can completely ignore our default and pass their own component to be rendered. This type also shouldn't be exported as it's internal only and not user-facing.

packages/hs-react-ui/src/components/Table/Table.tsx
donaldjbrady: I think that's a good idea. I've been trying to get as much customizability as I can and in doing so I think I'm going a little overboard. I was going for these use cases:

- Use the default icons that we have implemented (`mdiChevronDown` & `mdiChevronRight`)
- Use the default Icon component with different material ui components by using only `collapsedIcon` or `expandedIcon` as string props
- Use a completely expansion icon by using `expansionIconComponent` - completely replacing the material Icon we're using (`collapsedIcon` and `expandedIcon` are not required in this situation)
- Use a completely custom expansion icon component that also icon overrides. The use case I was thinking of here is an application that uses a custom expansion icon component that they've defined and can pass into any table, but they may want to customize the icon based on props instead of creating a whole new icon component for version. Essentially their own custom version of how our `ExpansionIcon` is implemented, but not requiring material-ui

I'd really like some input on this, can we talk about the approach and hash out what would be best and easiest for consumption without over complicating things?

packages/hs-react-ui/src/components/Table/Table.tsx
donaldjbrady: This is just the default Icon used. The implementation allows for a completely custom component instead that doesn't need to be material, but it does need to be React by using the `expansionIconComponent` prop :)

packages/hs-react-ui/src/components/Table/Table.tsx
aVileBroker: Ideally we should have one prop `ExpansionIcon`. We pass that component the "isCollapsed" info so it can decide how it wants to look/what icon it wants to return.

packages/hs-react-ui/src/components/Table/Table.stories.tsx
aVileBroker: If EmptyCell is just Table.Cell, couldn't you just pass that through?

aVileBroker: Oh nevermind - EmptyCell is a Table.Cell with nothing in it. cool.

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -151,31 +194,76 @@ const SortIcon = styled(Icon)`
   `}
 `;
 
+/** Start of variables */
+
+const defaultCollapsed: collapsedState = {};
+
+// Default expansion column added
+const collapsexpandedIconColumn = {
```
aVileBroker, line 202: I think we're missing an `e` here lol
```suggestion
const collapseExpandedIconColumn = {
```

packages/hs-react-ui/src/enums/colors.ts
```diff
@@ -7,6 +7,8 @@ enum colors {
   grayLight = '#7c8186',
   grayXlight = '#d3d6d7',
 
+  groupLabel = '#e8e8e8',
```
aVileBroker, line 10: This seems really specific - it should either named after a color descriptor or a more general use-case

Otherwise just pick a different gray or alter one from the constant file with polished (once we get custom color palletes, the table will still look right despite having a "custom" color)

donaldjbrady: In table I moved to grayXlight :)

packages/hs-react-ui/src/components/Table/Table.tsx
aVileBroker: Lit

PR #59, opened 2020-06-03
HSUI-34 Button enhancements

donaldjbrady:
Description of changes:

- Added `elevation` property and knob to the story
- Use drop shadow for all button types elevation shadow
- Added `Link` and `Outline` type buttons
- Added `color` prop that is used to change the color of the text of the Link and Outline type buttons as well as the outline for the Outline type Button. Also added knob for the color prop
- Added `fillColor` property used to select the background color of an `Outline` type button and the appropriate knob.
- Determine a background color for Outline type buttons when the color selected has poor contrast ratio with white. This can be overridden by using the newly added fillColor property.
- Updated `isLoading` property to be a boolean that determines whether or not to use a state with animation like the `Text` component
- Added `isProcessing` property and state like the `Text` component
- Added `iconPrefix` and `iconSuffix` properties used to prepend or append icons to the children
- Reorganized some of the contents of `Button.tsx`, `ButtonContainers.tsx`, and `Button.stories.tsx`

### DIFF
--- a/packages/hs-react-ui/src/components/Button/Button.stories.tsx
+++ b/packages/hs-react-ui/src/components/Button/Button.stories.tsx
@@ -9,6 +9,7 @@ import { withA11y } from '@storybook/addon-a11y';
 import { withDesign } from 'storybook-addon-designs';
 import Button from './Button';
 import fonts from '../../enums/fonts';
+import colors from '../../enums/colors';
 
 const ButtonTypes = Button.Types;
 const ButtonContainer = Button.Container;
@@ -31,22 +32,13 @@ storiesOf('Button', module)
   .add(
     'Basic Button',
     () => {
-      let useColor = false;
-      let useFillColor = false;
+      const useColor = boolean('Enable color property', false);
       const type = select('type', ButtonTypes, ButtonTypes.primary);
-
-      if (type === ButtonTypes.link || type === ButtonTypes.outline) {
-        useColor = boolean('Use color property', false);
-      }
-      if (type === ButtonTypes.outline) {
-        useFillColor = boolean('Use fillColor property', false);
-      }
 
       return (
         <Button
           type={type}
-          color={useColor ? color('color', '#000') : undefined}
-          fillColor={useFillColor ? color('fillColor', '#fff') : undefined}
+          color={useColor ? color('color', colors.grayDark) : undefined}
           onClick={action('button-click')}
           isLoading={boolean('isLoading', false)}
           elevation={number('elevation', 0)}
--- a/packages/hs-react-ui/src/components/Button/Button.stories.tsx
+++ b/packages/hs-react-ui/src/components/Button/Button.stories.tsx
@@ -32,13 +32,10 @@ storiesOf('Button', module)
   .add(
     'Basic Button',
     () => {
-      const useColor = boolean('Enable color property', false);
-      const type = select('type', ButtonTypes, ButtonTypes.primary);
-
       return (
         <Button
-          type={type}
-          color={useColor ? color('color', colors.grayDark) : undefined}
+          type={select('type', ButtonTypes, ButtonTypes.fill)}
+          color={color('color', colors.grayXlight)}
           onClick={action('button-click')}
           isLoading={boolean('isLoading', false)}
           elevation={number('elevation', 0)}
--- a/packages/hs-react-ui/src/components/Button/Button.stories.tsx
+++ b/packages/hs-react-ui/src/components/Button/Button.stories.tsx
@@ -10,9 +10,14 @@ import { withDesign } from 'storybook-addon-designs';
 import Button from './Button';
 import fonts from '../../enums/fonts';
 
-const ButtonContainers = Button.Containers;
 const ButtonTypes = Button.Types;
 const ButtonContainer = Button.Container;
+
+const options = {
+  none: '',
+  mdiLeadPencil,
+  mdiLoading,
+};
 
 addDecorator(withA11y);
 addDecorator(withDesign);
@@ -26,26 +31,38 @@ storiesOf('Button', module)
   .add(
     'Basic Button',
     () => {
-      const useColor = boolean('Use Color Prop', false);
-      return <Button
-        type={select('type', ButtonTypes, ButtonTypes.default)}
-        color={useColor ? color('color', '#000') : undefined}
-        onClick={action('button-click')}
-        isLoading={boolean('isLoading', false)}
-        elevation={number('elevation', 0)}
-      >
-        {text('children', 'Default text')}
-      </Button>
+      let useColor = false;
+      let useFillColor = false;
+      const type = select('type', ButtonTypes, ButtonTypes.primary);
+
+      if (type === ButtonTypes.link || type === ButtonTypes.outline) {
+        useColor = boolean('Use color property', false);
+      }
+      if (type === ButtonTypes.outline) {
+        useFillColor = boolean('Use fillColor property', false);
+      }
+
+      return (
+        <Button
+          type={type}
+          color={useColor ? color('color', '#000') : undefined}
+          fillColor={useFillColor ? color('fillColor', '#fff') : undefined}
+          onClick={action('button-click')}
+          isLoading={boolean('isLoading', false)}
+          elevation={number('elevation', 0)}
+          isProcessing={boolean('isProcessing', false)}
+          iconPrefix={select('iconPrefix', options, options.none)}
+          iconSuffix={select('iconSuffix', options, options.none)}
+        >
+          {text('children', 'Default text')}
+        </Button>
+      );
     },
     { design },
   )
   .add(
     'Themed Button',
     () => {
-      const isLoading = {
-        icon: <Icon path={mdiLoading} size={1} horizontal vertical rotate={90} />,
-        children: 'Loading',
-      };
       const icon = <Icon path={mdiLeadPencil} size={1} horizontal vertical rotate={90} />;
 
       const ThemedContainer = styled(ButtonContainer)`
@@ -70,8 +87,8 @@ storiesOf('Button', module)
       return (
         <Button
           StyledContainer={ThemedContainer}
-          isLoading={boolean('isLoading?', false) ? isLoading : null}
-          icon={boolean('Icon?', true) ? icon : null}
+          isLoading={boolean('isLoading', false)}
+          iconPrefix={boolean('iconPrefix', true) ? icon : undefined}
           onClick={action('button-click')}
         >
           {text('children', 'Edit')}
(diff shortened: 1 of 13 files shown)


packages/hs-react-ui/src/components/Button/Button.stories.tsx
```diff
@@ -22,25 +30,39 @@ const design = {
...
+      }
+      if (type === ButtonTypes.outline) {
+        useFillColor = boolean('Use fillColor property', false);
+      }
+
+      return (
+        <Button
+          type={type}
+          color={useColor ? color('color', '#000') : undefined}
+          fillColor={useFillColor ? color('fillColor', '#fff') : undefined}
```
aVileBroker, line 49: probably better to default to white on orange or darkGrey on lightGrey

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -1,7 +1,96 @@
...
+ */
+const getComplementaryColor = (color1: string, color2: string): string => {
+  const contrast = getContrast(color1, color2);
+
+  if (contrast > 4.5) {
+    return color2;
+  }
+
+  return complement(color1);
+};
```
aVileBroker, line 33: just use readableColor from polished :)
https://polished.js.org/docs/#readablecolor

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -1,7 +1,96 @@
...
+    background-size: 400% 100%;
+    width: 5rem;
+    height: 10px;
+    margin-top: -5px;
+    margin-bottom: -5px;
+    border-radius: 9999px;
+    animation: ${animation};
+    line-height: 0;
+  `}
+`;
```
aVileBroker, line 68: Should this be centralized? Or would it be less code to maintain if we just use a <Text> which has the processing prop set to true?

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -36,9 +125,41 @@ export const DestructiveButtonContainer = styled(ButtonContainer)`
...
+export const ButtonContainers: {
+  [index: string]: StyledComponentBase<any, {}, StyledButtonProps>;
+} = {
   [ButtonTypes.default]: ButtonContainer,
   [ButtonTypes.primary]: PrimaryButtonContainer,
   [ButtonTypes.secondary]: SecondaryButtonContainer,
   [ButtonTypes.destructive]: DestructiveButtonContainer,
+  [ButtonTypes.link]: LinkButtonContainer,
+  [ButtonTypes.outline]: OutlineButtonContainer,
 };
```
aVileBroker, line 165: Can we break from this pattern? I'd prefer just one container which can be easily overwritten - but which receives all the data for type and color as props

If that doesn't sound right lemme know your thoughts

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -1,7 +1,96 @@
...
+  fillColor?: string;
+};
+export enum ButtonTypes {
+  primary = 'primary',
+  secondary = 'secondary',
+  default = 'default',
+  destructive = 'destructive',
+  link = 'link',
+  outline = 'outline',
+}
```
aVileBroker, line 18: I think these should just be
- fill
- link
- outline

where primary, secondary, default, and destructive, are just accessable through choosing colors

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -1,7 +1,96 @@
...
-import { darken } from 'polished';
-import { ButtonTypes, ButtonContainer } from './Button';
+import styled, { StyledComponentBase, css, keyframes } from 'styled-components';
+import { darken, getContrast, complement } from 'polished';
 import colors from '../../enums/colors';
+import fonts from '../../enums/fonts';
+
+type StyledButtonProps = {
+  elevation: number;
+  color?: string;
```
aVileBroker, line 8: `textColor` here?

it's tough - I'd think that for outline and link, the border and text color would come from a `color` prop. And for "filled" we can automatically pick between white and (their dark color of choice) or black if their color of choice isn't dark enough; but that's too complicated.

It's tempting to just say "gimme a color and for `type: fill` we'll set the text color automatically"

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -1,66 +1,96 @@
...
-        {isLoading.children} {isLoading.icon}
-      </Text>
-    </StyledContainer>
+    <Container data-testid="hsui-button" elevation={elevation} color={color} fillColor={fillColor}>
+      <Progress />
+    </Container>
   ) : (
-    <StyledContainer onClick={onClick}>
+    <Container
+      data-testid="hsui-button"
```
aVileBroker, line 57: could this be `data-test-id`?

packages/hs-react-ui/src/components/Button/__tests__/Button.test.tsx
```diff
@@ -3,37 +3,47 @@ import { render, fireEvent, waitFor } from '@testing-library/react';
...
 import Button from '../Button';
 
+const testId = 'hsui-button';
+
 describe('Button', () => {
   it('fires click handler when clicked', () => {
     const clickSpy = jest.fn();
-    const { getByText } = render(<Button onClick={clickSpy}>Click me</Button>);
+    const { getByTestId } = render(
+      <Button data-testid={testId} onClick={clickSpy}>
```
aVileBroker, line 12: `data-test-id` here if you change it above

packages/hs-react-ui/src/components/Text/Text.tsx
```diff
@@ -43,11 +42,11 @@ const Progress = styled.div`
...
-        rgba(255, 255, 255, 0.2),
-        rgba(0, 0, 0, 0.2),
-        rgba(255, 255, 255, 0.2),
-        rgba(0, 0, 0, 0.2),
-        rgba(255, 255, 255, 0.2)
+        rgba(255, 255, 255, 0.75),
+        rgba(0, 0, 0, 0.75),
+        rgba(255, 255, 255, 0.75),
+        rgba(0, 0, 0, 0.75),
+        rgba(255, 255, 255, 0.75)
```
aVileBroker, line 49: what was the reason for this change?

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
donaldjbrady: I think this should be centralized.

Let me explain what was going on here and maybe you have some thoughts on the situation. I was trying to use the `Text` component for loading and all of the other states as well, but what ended up happening is that the button was constantly adding padding around the child elements that are not plain text (even spans). I replicated the changes in Button.tsx adding slightly negative margins to keep the button the same size regardless of state. Using Text, the button would jump sizes to at least an extra 16px tall.

packages/hs-react-ui/src/components/Button/Button.tsx
donaldjbrady: Yes, I'll do this. It's worth noting, however, that there is a very tiny configuration change required for the @testing-library/react: `configure({testIdAttribute: 'data-test-id'))` which may be required in other test files as well.

https://testing-library.com/docs/dom-testing-library/api-queries#overriding-data-testid

packages/hs-react-ui/src/components/Text/Text.tsx
donaldjbrady: I was trying to use the Text component inside of the button rather than reimplementing the loading style, but the opacity made the loading bar look really bad on some background colors because it bled through and seriously degraded the quality of the progress bar.

packages/hs-react-ui/src/components/Button/ButtonContainers.tsx
```diff
@@ -1,44 +1,86 @@
...
+  type: string;
+};
+export enum ButtonTypes {
+  primary = 'primary',
+  secondary = 'secondary',
+  default = 'default',
+  destructive = 'destructive',
+  link = 'link',
+  outline = 'outline',
```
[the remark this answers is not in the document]
aVileBroker: Maybe you could have it so that if type isn't in the list of normal types - you can try to see if they're trying to access a color in the color constants object? Otherwise default to fill I guess or throw angry error messages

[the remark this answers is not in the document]
aVileBroker: For me, `type` should more about how the color they specified is applied, and the general shape of the button. Using type to control color when we already expose a color prop will make it confusing as to which has priority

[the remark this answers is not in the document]
donaldjbrady: I went ahead and removed primary/secondary/default/destructive and replace it with fill. Without type provided, the button defaults to fill. Without color provided, the button defaults to the old 'default' button that is lightgray background with dark gray text

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -1,66 +1,104 @@
...
   StyledContainer = ButtonContainer,
-  icon,
+  iconPrefix,
+  iconSuffix,
   isLoading,
+  isProcessing,
   children,
+  elevation = 0,
+  type = ButtonTypes.fill,
+  color,
```
aVileBroker, line 54: no default color?

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -1,66 +1,104 @@
...
+  LoadingBar = StyledProgress,
 }: ButtonProps) => {
+  const Container = StyledContainer || ButtonContainer;
   return isLoading ? (
-    <StyledContainer>
-      <Text>
-        {isLoading.children} {isLoading.icon}
-      </Text>
-    </StyledContainer>
+    <Container data-test-id="hsui-button" elevation={elevation} color={color} type={type}>
```
aVileBroker, line 60: I think it should still have an onClick - if they don't want to pass it until loading is false that's on them - but we shouldn't force that business logic on them

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -1,66 +1,104 @@
 import React from 'react';
+import UnstyledIcon from '@mdi/react';
+import { mdiLoading } from '@mdi/js';
 import styled, { StyledComponentBase } from 'styled-components';
-import colors from '../../enums/colors';
-import fonts from '../../enums/fonts';
+import ButtonContainer, { ButtonTypes, ButtonContainerProps } from './ButtonContainers';
```
aVileBroker, line 5:
```suggestion
import colors from '../../enums/colors';

import ButtonContainer, { ButtonTypes, ButtonContainerProps } from './ButtonContainers';
```

PR #60, opened 2020-06-04
Add Dropdown component

joien1:


packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -35,7 +33,7 @@ export type ButtonProps = {
   StyledContainer?: string & StyledComponentBase<any, {}>;
   icon?: any;
   isLoading?: any;
-  children?: string | Node;
+  children?: ReactNode;
```
aVileBroker, line 36: I've been just putting `any` here since we could potentially let them pass in an array or ReactNodes as well

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { withDesign } from 'storybook-addon-designs';
+import { Dropdown } from './Dropdown';
+
+addDecorator(withDesign);
+const design = {
+    type: 'figma',
+    url: 'https://www.figma.com/file/3r2G00brulOwr9j7F6JF59/Generic-UI-Style?node-id=102%3A28',
+};
+
+const Background = styled.div`background: white; height: 100%;`;
```
aVileBroker, line 16: Have you pulled from master? I feel like this isn't necessary anymore.

But either way it stops people from being able to use the storybook background feature to see what the dropdown looks like on a dark background

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+}}
+  
+`;
+const ValueContainer = styled(ButtonContainer)`
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  align-items: center;
+  
+  line-height: initial;
```
aVileBroker, line 30: does it really break anything to put this change into `Button`?

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+  width: 15rem;
```
aVileBroker, line 31: could you add a TODO here to add constants for widths - like `fullWidth: 100%` and `inputWidth: 15rem`?

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+  background: white;
+  position: absolute;
+  max-height: 10rem;
+  overflow-y: scroll;
+  width: 15rem;
+  border: 0.5px solid ${colors.blueCharcoal25};
+`;
+
+const OptionItem = styled.div`
+  height: 2rem;
```
aVileBroker, line 57: does this look funky when there's two lines of text? Maybe just padding here would be better

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+  display: flex;
+  align-items: center;
+  
+  font-family: Montserrat,Roboto,sans-serif;
```
aVileBroker, line 61: I feel like this should be on the container - and maybe just be `{fonts.body}` to make sure it gets imported when this is the only component brought in

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+  clearable?: boolean,
+  elevation?: number,
+  multi?: boolean,
+  name: string,
+  onClear?: () => void,
+  onSelect?: (selected: string | Array<string>) => void,
+  options: Array<string>,
+  tabIndex?: number,
+  values?: Array<string>,
```
aVileBroker, line 94: for my own education, what's the difference between this and `string[]`?

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+  });
+
+  const handleBlur = useCallback((e: React.FocusEvent) => {
+    e.preventDefault();
+    const target = e.nativeEvent.relatedTarget as HTMLElement | null;
+    // check if we're focusing on something we don't control
+    if (!target || (target.id && !target.id.startsWith(state.id))) {
+      setState(curState => ({ ...curState, isOpen: false }));
+    }
+  }, [state.id]);
```
aVileBroker, line 128: We should probably add an onBlur callback to go with this guy

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+      >
+        <StyledValueItem>{((values.length && values) || state.selectedValues).join(', ')}</StyledValueItem>
+        <ValueIconContainer>
+          {clearable && (
+            <Icon
+              // TODO Fix issue with Icon not correctly exposing its onClick on its declaration file
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              onClick={handleClear}
+              onFocus={(e: FocusEvent) => e.stopPropagation()}
```
aVileBroker, line 239: The mdi docs say they don't want icon to have onclick for whatever reason - But you could wrap it in a div with those attached to it. (we probably want the padding around the icon to be clickable as well really)

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,279 @@
...
+            >
+              <CheckContainer>
+                {state.selectedValues.includes(opt)
+                  && (
+<Icon
+  path={mdiCheck}
+  size={0.75}
+  color={readableColor(colors.blueCharcoal50, colors.success)}
+/>
+)
```
aVileBroker, line 268: this indention lookin real weird on gh

packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -470,8 +470,7 @@ const Table = ({
             >
               {Object.keys(copiedColumns).map(headerColumnKey => {
                 const RenderedCell = copiedColumns[headerColumnKey].cellComponent || StyledCell;
-                const breakPointHit =
-                  width > (copiedColumns[headerColumnKey].minTableWidth || Infinity);
+                const breakPointHit =                  width > (copiedColumns[headerColumnKey].minTableWidth || Infinity);
```
aVileBroker, line 473: not sure if prettier is drunk

packages/hs-react-ui/src/components/Button/Button.tsx
joien1: I believe ReactNode covers everything:
https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/cc2ec67e0b36b7e838df8f2a54b4cd84df6c78cd/README.md#useful-react-prop-type-examples

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
joien1: Nope I forgot to remove this.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Doesn't seem like it. Generally I am trying to avoid modifying components (*goes on to modify Button), but maybe I should not have that mindset right now.

joien1: Made the change to button and seems to not affect anything.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: It does look funky - made change.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Good point.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: https://stackoverflow.com/questions/38239579/what-is-the-difference-arraystring-and-string

TL;DR Nothing, except Array<string> can be extended, string[] can't - that's why I just tend to prefer using Array<string>

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Interesting, I wonder why that's the case. I've moved it into a div.

packages/hs-react-ui/src/components/Table/Table.tsx
joien1: It's drunk. I'll fix it >.<

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Done.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Done.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Done.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: Prettier had a few too many.

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
aVileBroker: neat

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+storiesOf('Dropdown', module).add(
+  'Basic',
+  () => {
+    const options = ['Apple', 'Banana', 'Grapefruit', 'Grape', 'Orange', 'Watermelon'];
+    return (
+      <>
+        <label htmlFor="fave-fruit">What's your favorite fruit?</label>
+        <Dropdown
```
aVileBroker, line 22: one thing that concerns me is that the container should be the source of truth

joien1: Can you elaborate on that?

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+          clearable={boolean('clearable', false)}
+          elevation={number('elevation', 2, { range: true, min: 0, max: 5, step: 1 })}
+          multi={boolean('multi', false)}
+          name="fave-fruit"
+          onBlur={action('onBlur')}
+          onClear={action('onClear')}
+          onSelect={action('onSelect')}
+          options={options}
```
aVileBroker, line 30: should we include a `values` knob? It'd just be a select with the values - and we wouldn't be able show multiselects properly but idk. May be better to put into the docs

packages/hs-react-ui/src/enums/colors.ts
```diff
@@ -9,6 +9,11 @@ enum colors {
 
   success = '#698F00',
   destructive = '#d00',
+
+  blueCharcoal25 = 'rgba(211, 214, 215, 0.25)',
+  blueCharcoal50 = 'rgba(211, 214, 215, 0.5)',
+  blueCharcoal75 = 'rgba(211, 214, 215, 0.75)',
+  blueCharcoal100 = 'rgba(211, 214, 215, 1)',
```
aVileBroker, line 16: these are grayDark with different opacities, not the headstorm colors that match these. We also want the users to be able to change these eventually and calling it "blue" and "charcoal" will be weird when they use a warm grey base.

I vote grayDark25, grayDark50, etc

joien1: Sure - there's a comment in figma's styles above the color I used that said blueCharcoal, which is why I used that name here. Maybe figma should also be changed.

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
joien1: I've included one (doesn't work for multi-select), but yeah we'll definitely want to put this in the docs anyway.

packages/hs-react-ui/src/enums/colors.ts
aVileBroker: Sounds good to me

PR #64, opened 2020-06-09
Add box-sizing and font-family to all base HTML elements used

joien1:
Devs should now use these elements via `styled(Div)` rather than `styled.div`

### DIFF
--- a/packages/hs-react-ui/src/htmlElements/index.tsx
+++ b/packages/hs-react-ui/src/htmlElements/index.tsx
@@ -1,7 +1,8 @@
+import { ComponentType } from 'react'
 import styled, { StyledComponent } from 'styled-components';
 import fonts from '../enums/fonts';
 
-const withGlobalStyle = (Component: StyledComponent<any, any>) => styled(Component)`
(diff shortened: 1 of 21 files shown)


packages/hs-react-ui/src/htmlElements/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const Div = withGlobalStyle(styled.div``);
+export const Span = withGlobalStyle(styled.span``);
+export const Button = withGlobalStyle(styled.button``);
+export const Input = withGlobalStyle(styled.input``);
+export const Label = withGlobalStyle(styled.label``);
+export const HR = withGlobalStyle(styled.hr``);
+export const Table = withGlobalStyle(styled.table``);
+export const TH = withGlobalStyle(styled.th``);
+export const TD = withGlobalStyle(styled.td``);
+export const TR = withGlobalStyle(styled.tr``);
```
donaldjbrady, line 21: Can you add some info to the README that gives context about this change to onboarding developers so that they have a place they should be seeing to use these HOC's?

joien1: Great point - I've made an update to the README.

PR #66, opened 2020-06-11
Modal improvements

rfisher58:


packages/hs-react-ui/src/components/Modal/Modal.tsx
```diff
@@ -34,7 +36,7 @@ const ModalUnderlay = styled(Div)<{ backgroundBlur: string; backgroundDarkness:
 
     z-index: 1000;
 
-    backdrop-filter: blur(${backgroundBlur}) opacity(${1 - backgroundDarkness / 100});
+    backdrop-filter: blur(${backgroundBlur}) brightness(${backgroundDarkness});
```
aVileBroker, line 39: darkness is the inverse of brightness, so darkness 1 sounds like MAX darkness, but it's actually the opposite.

You could rename the prop to background brightness, or just subtract by 1 and keep the prop name (which is my suggestion)
```suggestion
    backdrop-filter: blur(${backgroundBlur}) brightness(${1 - backgroundDarkness});
```

aVileBroker:
Comments copied from slack
- change insideClose to insetCloseButton
- change modalCloseButton to be a styled Button (from hs-ui)
- position should always be absolute
- if it's inset it has top: 2rem, right: 2rem
- if it's not inset it has top -4rem, right: -4rem - and it should be white with an elevation of 1 or 2
- Delete the card and the associated props, move that to the story
- Move the button into the StyledContainer, after the {children}

PR #69, opened 2020-06-12
HSUI 42, 62, 60, 63 Example Form

donaldjbrady:
* Created an example of a simple controlled form using all HSUI elements except for Table.
* Excluded the example form from unit test coverage
* 60 - Fixed Text input's input not growing to container size
* 62 - Fixed Button icons increasing button height by 1pixel shifting borders of other components
* 63 - Vertically centered icons in Text with the text

### DIFF
--- a/packages/hs-react-ui/src/components/Button/Button.tsx
+++ b/packages/hs-react-ui/src/components/Button/Button.tsx
@@ -80,7 +80,7 @@ export const ButtonContainer: string & StyledComponentBase<any, {}, ButtonContai
 
     return `
       display: inline-flex;
-      font-size: 1rem;
+      font-size: 1em;
       padding: .75em 1em;
       border-radius: 0.25em;
       transition: filter ${timings.slow};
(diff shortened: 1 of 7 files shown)


packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -79,8 +79,8 @@ export const ButtonContainer: string & StyledComponentBase<any, {}, ButtonContai
     const shadowOpacity = 0.5 - elevation * 0.075;
 
     return `
-      display: inline-block;
-      font-size: 1em;
+      display: inline-flex;
+      font-size: 1rem;
```
aVileBroker, line 83: Kinda 50/50 on this one - but buttons should probably fit into the text around them
```suggestion
      font-size: 1em;
```

packages/hs-react-ui/src/components/Text/Text.tsx
```diff
@@ -38,7 +38,7 @@ const IconContainer = styled(Span)`
   ${({ side }: { side: 'left' | 'right' }) => `
     margin-${side === 'left' ? 'right' : 'left'}: .5em;
     display: inline-flex;
-    vertical-align: text-bottom;
+    vertical-align: middle;
```
aVileBroker, line 41: I kinda liked it text-bottom but maybe it looks weird in actual usage

PR #73, opened 2020-06-16
Hsui 55 fix zindex

enyiasonye:


packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -69,6 +70,8 @@ const OptionsContainer = styled(Div)`
   top: 100%;
   left: 0px;
   max-height: 10rem;
+  top: 100%;
+  left: 0px;
```
aVileBroker, line 74: oops it was already there lol

PR #75, opened 2020-06-16
Range slider

aVileBroker:
Feedback welcome on structure, types, everything :D (even the name)


joien1:
Did you mean to include the package.lock file?

packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,312 @@
...
+    transition: filter .1s;
+
+    ${
+      disabled
+        ? `
+      filter: grayscale(1) contrast(.5) brightness(1.2);
+      pointer-events: none;
+    `
+        : ''
+    }
```
joien1, line 45: Not sure if prettier went ham, but there's a lot of unneeded line breaks here.

packages/hs-react-ui/src/components/RangeSlider/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  motionBlur?: boolean;
+  springOnRelease?: boolean;
+  debounceInterval?: number;
+  axisLock?: 'x' | 'y' | '';
+  onDrag?: (val) => void;
+  disabled?: boolean;
+  min: number;
+  max: number;
+  values?: number[] | valueProp[];
+  testId: string;
```
joien1, line 45: Imo should be optional, maybe with a sensible default

packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
aVileBroker: It really went crazy

packages/hs-react-ui/src/components/RangeSlider/types.ts
aVileBroker: oops yes

aVileBroker: I don't think it needs a default as this just gets appended to the default... maybe it should be called testIdSuffix or something lol

PR #81, opened 2020-07-28
HSUI-121 moved storybook deps 

ericbaumann:


joien1:
We need to do a bit more work than just this so we'll close this one.

PR #85, opened 2020-08-03
Add FoundryProvider to allow for customization of global stylesheet.

joien1:


packages/hs-react-ui/src/context/index.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useContext } from 'react';
+import styled, { css, StyledComponentBase } from 'styled-components';
+import fonts from '../enums/fonts';
+import colorsEnum from '../enums/colors';
+
+export const defaultGlobalStyles = `
+  ${
+    process.env.NODE_ENV !== 'test'
```
donaldjbrady, line 8: Are we polyfilling process.env so that this doesn't throw an error when running in a browser not through storybook? I'm only asking because I'm not sure how storybook rendering works or if it does any server side rendering for us.

joien1: I've updated the logic for just in case, but it should work regardless. :D

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -60,12 +67,17 @@ export const getFontColorFromVariant = (variant: string, color: string) => {
...
+export const getBackgroundColorFromVariant = (
+  variant: string,
+  color: string,
+  transparentColor: string,
+) => {
   switch (variant) {
     case ButtonVariants.text:
     case ButtonVariants.outline:
-      return colors.transparent;
+      return transparentColor;
```
aVileBroker, line 80: if it's optional, should it be more like
```suggestion
      return transparentColor || 'transparent';
```

Couldn't do the same for color as we don't have a sensible default we can recommend, and transparent is brand neutral so prob safe. Should probably make color and variant required in the jsdoc/typing

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -75,8 +87,13 @@ export const ButtonContainer: string & StyledComponentBase<any, {}, ButtonContai
   ButtonElement,
 )`
   ${({ elevation = 0, color, variant }: ButtonContainerProps) => {
-    const backgroundColor = getBackgroundColorFromVariant(variant, color);
-    const fontColor = getFontColorFromVariant(variant, color);
+    const { transparent, background, grayDark } = useColors([
+      'transparent',
+      'background',
+      'grayDark',
+    ]);
```
aVileBroker, line 94: making this an array is a bit confusing to me, and they wouldn't be able to get intellisense stuff to define what color array should be passed right?

aVileBroker: are the keys being passed in just like requests for those items? But then we're destructuring so I feel like we could just have useColors return the entire colorscheme object

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -147,18 +162,20 @@ const Button = ({
...
   variant = ButtonVariants.fill,
   type = ButtonTypes.button,
-  color = colors.grayDark,
+  color,
   onClick,
   LoadingBar = StyledProgress,
 }: ButtonProps): JSX.Element | null => {
   const hasContent = Boolean(children);
+  const grayDark = useColors('grayDark');
+  const containerColor = color || grayDark;
```
aVileBroker, line 171: Choosing dark grey for their default makes sense but is definitely a stylistic choice... but the only other option would be to make color required which isn't "fun for the whole family right out of the box" so I guess this is best :-/

packages/hs-react-ui/src/components/Button/Button.tsx
joien1: We do get intellisense from the typescript overloads (I pulled out all the stops for this one), but as I was working I realized that you could just destructure the colors object. So, I'll just remove the other options and go with the destructure option across the board.

packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -147,18 +162,20 @@ const Button = ({
   elevation = 0,
   variant = ButtonVariants.fill,
   type = ButtonTypes.button,
-  color = colors.grayDark,
+  color,
   onClick,
   LoadingBar = StyledProgress,
 }: ButtonProps): JSX.Element | null => {
   const hasContent = Boolean(children);
+  const grayDark = useColors('grayDark');
```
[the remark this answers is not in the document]
joien1: Addressed above, going to move to simple destructuring.

packages/hs-react-ui/src/components/Button/Button.tsx
joien1: I didn't make any opinionated changes in this PR. This is making sure to handle the old default case on color from the props. 

Totally agree with you, the dark button seems weird. These types of concerns seem like something where we should put up all of our components on one page with the most minimal configuration possible and see what sticks out and address them there.

packages/hs-react-ui/src/components/Button/Button.tsx
joien1: Change made

packages/hs-react-ui/src/components/Button/Button.tsx
aVileBroker: agreeb

packages/hs-react-ui/src/components/Modal/Modal.tsx
```diff
@@ -97,46 +97,49 @@ const Modal = ({
...
           StyledContainer={StyledCloseButtonContainer}
           containerProps={{
             closeButtonAttachment,
           }}
           iconPrefix={mdiClose}
-          color={closeButtonAttachment === 'inside' ? colors.grayDark : colors.background}
-          elevation={closeButtonAttachment === 'inside' ? 0 : 1}
+          color={colors.background}
+          elevation={1}
           type="link"
```
aVileBroker, line 113: I don't think these changes are right - they should still be in the ternary. The issue is that `type` was changed to `variant` but this file didn't get updated so it is wrong. 
```suggestion
          color={closeButtonAttachment === 'inside' ? colors.grayDark : colors.background}
         elevation={closeButtonAttachment === 'inside' ? 0 : 1}
          variant="text"
```

aVileBroker: Oh you split it out so that they're different branches

packages/hs-react-ui/src/components/Modal/Modal.tsx
```diff
@@ -97,46 +97,49 @@ const Modal = ({
...
+        {closeButtonAttachment !== 'corner' && (
+          <StyledCloseButton
+            StyledContainer={StyledCloseButtonContainer}
+            containerProps={{
+              closeButtonAttachment,
+            }}
+            iconPrefix={mdiClose}
+            color={closeButtonAttachment === 'inside' ? colors.grayDark : colors.background}
+            elevation={closeButtonAttachment === 'inside' ? 0 : 1}
+            type="link"
```
aVileBroker, line 129:
```suggestion
            variant="text"
```

PR #86, opened 2020-08-04
FOUNDRYUI-64 Close dropdown when clicking the value -> v1-release

donaldjbrady:
Note: Still need the package.json bump


packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -37,6 +37,8 @@ export type ButtonProps = {
   type?: ButtonTypes;
   color?: string;
   onClick: (...args: any[]) => void;
+  onMouseDown?: (...args: any[]) => void;
+  onMouseUp?: (...args: any[]) => void;
```
joien1, line 41: I think this can be `onMouseDown?: (e?: React.MouseEvent)`

PR #87, opened 2020-08-04
FOUNDRYUI-83 add placeholder prop to Dropdown

donaldjbrady:
Note: Using Text to display the placeholder instead of TextInput because it does not have a disabled prop.


packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -198,6 +208,12 @@ const Dropdown = ({
 }: DropdownProps): JSX.Element | null => {
   const [isOpen, setIsOpen] = useState<boolean>(false);
 
+  // Merge the default styled continer prop and the placeholderProps object to get user styles
```
joien1, line 211:
```suggestion
  // Merge the default styled container prop and the placeholderProps object to get user styles
```

packages/hs-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -198,6 +208,12 @@ const Dropdown = ({
...
+  const placeholderMergedProps = {
+    ...{ StyledContainer: PlaceholderContainer },
+    ...placeholderProps,
+  };
```
aVileBroker, line 215: Should this be the StyledValueContainer we're spreading?

donaldjbrady: No, this is for the placeholder specifically. The placeholder uses the Text component and this is to apply the default styling and override that by any of the placeholderProps that get passed in. I figured for most use cases they wouldn't want to customize the placeholder, but it's easy enough to do with the placeholderProps prop.

PR #94, opened 2020-08-05
FOUNDRYUI-97 Move shadow calculations to single function -> master

donaldjbrady:


packages/hs-react-ui/src/utils/styles.ts
```diff
@@ -0,0 +1,30 @@
+// A constant factor that works well with base 10 logarithms
+const elevationFactor = 10 ** 0.1;
+
+/**
+ * Calculates the values for shadows based on the provided elevation
+ * @param {number} elevation - The elevation to get values for
+ */
```
aVileBroker, line 7: jsdoc should probably document the return

packages/hs-react-ui/src/utils/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  // Using abs will return the same values for negative and positive elevations
+  const logVal = Math.log10(Math.abs(elevation) + elevationFactor);
+
+  // Set the updated values for the elevation
+  elevationValues.xOffset = 0;
+  elevationValues.yOffset = (logVal / 4) * 0.5;
+  elevationValues.blur = logVal * 0.5;
+  elevationValues.opacity = 0.5 - logVal * 0.2;
+  return elevationValues;
```
aVileBroker, line 29: I'm in two minds about this one - so far all the shadows you updated used the same string with these variables.. should we just return the full shadow string?

The advantage of that approach would mean that we could switch to a box shadow when it's negative and assume that if they want an inset shadow that they're okay with a box-shadow?

donaldjbrady: I was thinking the same thing honestly. I almost did that, but I wasn't sure that made complete sense, but now that you've also mentioned it as well I feel a lot better about returning a string and will update it accordingly (including box-shadow for negative values!)

packages/hs-react-ui/src/utils/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+  console.log(`Elevation: ${elevation}`, elevationValues);
+  return elevationValues;
+};
+
+/**
+ * Returns the shadow style for a given elevation. If the elevation is negative, box-shadow is used and is inset.
+ * @param {number} elevation
+ * @returns {string} The css style property and value
+ */
+export const getElevationShadowStyle = (elevation = 0) => {
```
aVileBroker, line 50: Propose calling it `getShadowStyleFromElevation` or something similar

or it could be`getShadowStyle` to be simpler (this assumes they'd look at the parameters the function takes or jsdoc to understand it)

packages/hs-react-ui/src/utils/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+  const logVal = Math.log10(absVal + elevationFactor);
+
+  // Set the updated values for the elevation
+  elevationValues.xOffset = 0;
+  elevationValues.yOffset = isNegative
+    ? Math.round(Math.E ** (absVal / 10) + 0.1) / 16
+    : calc() / 16;
+  elevationValues.blur = isNegative ? (absVal * 2) / 16 : absVal * 0.25;
+  elevationValues.opacity = /** isNegative ? 0.3 - 0.05 * absVal : */ 0.5 - logVal * 0.2;
+  console.log(`Elevation: ${elevation}`, elevationValues);
```
aVileBroker, line 41: console?

donaldjbrady: Oof - used that during debugging and missed that log 🤦‍♂️ Good catch

packages/hs-react-ui/src/utils/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+export const getShadowStyle = (elevation = 0) => {
+  if (elevation === 0) {
+    return '';
+  }
+
+  const { xOffset, yOffset, blur, opacity } = calculateElevationValues(elevation);
+
+  return elevation > 0
+    ? `filter: drop-shadow(${xOffset}rem ${yOffset}rem ${blur}rem rgba(39, 47, 78,${opacity}));`
+    : `box-shadow: inset ${xOffset}rem ${yOffset}rem ${blur}rem rgba(39, 47, 78,${opacity});`;
```
aVileBroker, line 59: We should probably use polished to break apart a color from the theme into rgb and recombine them in this string template to set the opacity

Only problem with this is that we don't have access to the providers color - only the constant object. Unless we accept an optional shadow color prop for this which would still allow it to be overwritten.

PR #96, opened 2020-08-05
Add InteractionFeedback component -> master

joien1:


joien1:
Old comments on https://github.com/Headstorm/rasa-ui/pull/67

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,97 @@
+import React, { useCallback, useState } from 'react';
+import { animated, useTransition } from 'react-spring';
+import styled, { StyledComponentBase } from 'styled-components';
+import shortid from 'shortid';
+import colors from '../../enums/colors';
+import useResizeObserver from 'use-resize-observer/polyfilled';
+
+// TODO Reduce amount of any/ts-ignore done here
```
aVileBroker, line 8: I don't see any - is this TODO still relavent?

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const Circle = styled.svg`
```
aVileBroker, line 10: Propose naming `Circle` to `SVGContainer` or something similar, expose it as a subcomponent, and then rename `fragment` to `shape`
```suggestion
const SvgContainer = styled.svg`
```

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,97 @@
...
+}: InteractionFeedbackProps) => {
+  const { ref, width, height } = useResizeObserver();
+  const [animations, setAnimations] = useState<Array<Animation>>([]);
+
+  const transitions = useTransition<Animation, Animation>(animations, {
+    keys: (item: Animation) => item.id,
+    onRest: (item: Transition) => setAnimations(a => a.filter(ani => ani.id === item.id)),
+    ...transitionProps,
+  });
+  const fragment = transitions((style, item) => {
```
aVileBroker, line 50:
```suggestion
  const circles = transitions((style, item) => {
```

aVileBroker:
Thoughts on moving from transitioning the `r` and going with a `transform` which animates `scale(0)` to `scale(1)` (where the circle starts at the `max(height, width)` of the component)?

I guess we can make it a change later, but being tied to a circle would make this an exception to the interchangeable subcomponent philosophy. Of course we would have to enforce that they use something from the `animated` object, like circle, path, g, etc - so I could see us just taking a string which relates to an svg element which we use like `animated[shapeElement]` or something

packages/hs-react-ui/package.json
```diff
@@ -88,13 +88,15 @@
   "dependencies": {
     "@mdi/js": "^5.2.45",
     "@mdi/react": "^1.4.0",
+    "@types/shortid": "^0.0.29",
```
donaldjbrady, line 91: Very minor, but types are usually dev dependencies

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,97 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+`;
+
+const Container = styled(animated.div)`
+  position: relative;
+`;
+
+enum feedbackTypes {
```
donaldjbrady, line 21: It may be worthwhile to move this to the enums directory and import it to keep consistent with other components?

aVileBroker: Do these feedback types do anything at the moment? I figure giving them the ability to directly set transition config control negates the need for presets.. Though I could see it being a thing in the future perhaps

packages/hs-react-ui/package.json
joien1: Good catch, thanks.

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
joien1: Doesn't seem so, removed.

PR #105, opened 2020-08-10
CI integration -> master

donaldjbrady:
Description of Changes:
- Updated root package.json scripts to be smaller units of work
- Updated react package.json scripts to run in a less redundant way - no more prettier and builds on every command!
- Add husky and commitizen for commit hooks and commit message formatting for release notes
- Add semantic-release
- Add semantic-release changelog plugin to generate changelogs
- Collect coverage and commit on push to release branches
- Update storybook docs on push to release branches
- Add package.json script entry for making commits formatted for the changelog


.github/workflows/buildCheck.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Prepare
+        run: npm install
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Lint React
+        run: npm run lint-react
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
joien1, line 33: Are these tokens needed in the steps where we're not installing anything?

donaldjbrady: Right now they are not, but I put them there since they don't hurt anything. The advantage of having them there now is less work to do for CICD if we end up depending on our own private package.

packages/hs-react-ui/package.json
```diff
@@ -118,7 +122,47 @@
     "clean": "rm -rf ./dist/*",
     "webpack": "webpack --config webpack.config.js",
     "dts": "tsc --declaration --outDir dist/types/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false",
-    "test": "jest --collectCoverage=true"
+    "pre-push": "yarn run lint-ci && yarn run test && yarn run clean && yarn run webpack && yarn run dts",
+    "test": "rm -rf ./coverage/* && jest --collectCoverage=true",
```
joien1, line 126: Does jest not do this already? (removing /coverage directory) Bummer.

packages/hs-react-ui/package.json
```diff
@@ -118,7 +122,47 @@
...
+        "path": "@semantic-release/git",
+        "assets": [
+          "package.json",
+          "../../docs/**",
+          "./coverage/**",
+          "CHANGELOG.md"
+        ],
+        "message": "chore(release): ${nextRelease.version} [skip ci]nn${nextRelease.notes}"
+      }
+    ]
```
joien1, line 165: This is going to make the versions auto increment without us needing to do that, is that right?

packages/hs-react-ui/package.json
donaldjbrady: I thought it was supposed to, but it didn't look like it was. It DOES overwrite the old ones, but for consistency sake across the scripts for generating the storybook docs and the coverage info

packages/hs-react-ui/package.json
donaldjbrady: Yes, it will! Semantic release will auto version based on commit messages

PR #106, opened 2020-08-10
Interaction feedback -> v1-release

joien1:
As we decided in standup, this is a checkpoint for the InteractionFeedback component. I'll get a ticket made to address feedback left in https://github.com/Headstorm/rasa-ui/pull/96, namely that we can't customize the component as much as we'd like to. 

One change addressed from prior comments is removing the feedbackTypes enum, as that's not relevant to the current usage of the component.


packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,89 @@
...
+`;
+
+type Animation = { cx: string; cy: string; id: string };
+type Transition = { r: string } & Animation;
+type InteractionFeedbackProps = {
+  StyledContainer?: string & StyledComponentBase<any, {}>;
+
+  children: React.ReactNode;
+  interpolationFunctions?: Record<string, (val: any) => any>;
+  transitionProps: any;
```
donaldjbrady, line 26: If we're going to default these values, transitionProps can be optional too

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const InteractionFeedback = ({
+  StyledContainer = Container,
+
+  children,
+  interpolationFunctions = {},
+  transitionProps,
```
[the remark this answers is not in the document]
joien1: good point. I've defaulted them to the same values as the demo and attached them to the InteractionFeedback function like we do for StyledContainers

packages/hs-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
joien1: totes. Also added a comment saying to make types gud

PR #110, opened 2020-08-10
Add disabled props to Button, TextInput, Checkbox

joien1:
This'll go after the Checkbox variants PR


packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -130,35 +138,25 @@ const Button = ({
...
+    id,
+    onClick,
+    onMouseDown,
+    onMouseUp,
+    elevation,
+    color: containerColor,
+    variant,
+    type,
+    disabled,
+  };
```
aVileBroker, line 152: Missing containerProps? Also probably needs a more descriptive name than just "props"

joien1: You right

PR #117, opened 2020-08-11
FOUNDRYUI-136 Add DocsPage args table to stories

wmangimelli:
Screen Shot 2020-08-11 at 11 20 49 AM

### DIFF
--- a/packages/hs-react-ui/src/components/Button/Button.tsx
+++ b/packages/hs-react-ui/src/components/Button/Button.tsx
@@ -30,18 +30,57 @@ export enum ButtonTypes {
 }
 
 export type ButtonProps = {
+    /**
+  ButtonContainer
+  */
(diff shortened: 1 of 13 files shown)


packages/hs-react-ui/src/components/Button/Button.tsx
```diff
@@ -30,18 +30,57 @@ export enum ButtonTypes {
 }
 
 export type ButtonProps = {
+    /**
```
donaldjbrady, line 33: Are we going to add these types of comments to all of the props for each of the components for the docs to be created?

joien1: We will need to do this, but this should be a separate task imo. @wmangimelli could you make a ticket on the Foundry board for this task?

wmangimelli: Yes, will do.

PR #124, opened 2020-08-17
Feature/foundryui 79 Click range slider to select values

donaldjbrady:


packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -233,6 +234,54 @@ export default ({
...
+      });
+
+      if (closestVal) {
+        // TODO: use the closest val to find the handle to move and move it
+      }
+
+      // Set the slider position - this causes issues with the story where the value is the rounded to the same
+      // previous value. Clicking on 2.3 when set to 2 will move the handle, but the value doesn't get updated via
+      // the values prop, seems like a deep compare of the properties is happening on render for the values prop
+      set({
```
donaldjbrady, line 269: @aVileBroker I wasn't sure if we should be calling `set` or not. With how the story for the component is written to round the value, clicking on the rail twice at ~2.3 will first move the handle and fill the range slider to 2 and then the second click will move the handle, but the rail fill will remain umoved. If we do not use the `set` function, the handle moves and the bar moves as expected, would it be best to remove the set and expect the handles to always be set based off of the incoming values?

aVileBroker: It looks like the animation isn't working currently anyway - you can set "springOnRelease" to `true` and it doesn't actually do the set call to animate the handle - it just snaps to the old position. The set is mainly used to animate the handle smoothly to the rounded position.

joien1:
Are we good to merge this one?

PR #125, opened 2020-08-17
59 TextInput extends native input props

wmangimelli:


packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -194,6 +195,7 @@ const TextInput = ({
         id={id}
         type={type}
         multiLineIsResizable={multiLineIsResizable}
+        {...rest}
```
joien1, line 198: @aVileBroker @wmangimelli What do you think about removing all of the props like `type` as they are implicitly included? 

I'd also recommend renaming `rest` to something more descriptive. Right now, rest refers to all of the React HTML input element props.

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -176,24 +154,15 @@ const TextInput = ({
       <InputComponent
         cols={cols}
         rows={rows}
-        aria-label={ariaLabel}
-        placeholder={placeholder}
         onChange={(e: SyntheticEvent) => {
           e.persist();
           onChange(e);
```
aVileBroker, line 159: looks like onChange is removed above, this should pull from the rest object
```suggestion
          nativeInputHTMLAttributes.onChange(e);
```

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -176,24 +154,15 @@ const TextInput = ({
...
           debouncedChange(e);
         }}
-        onKeyPress={onKeyPress}
-        onKeyDown={onKeyDown}
-        onKeyUp={onKeyUp}
-        onFocus={onFocus}
-        onBlur={onBlur}
-        onReset={onReset}
-        onInput={onInput}
         value={value || defaultValue}
```
aVileBroker, line 162: We shouldn't do this anymore - they'll pass either value or default value and we should respect that and not put both into value

wmangimelli: Gotcha, I am going to remove them altogether so they are just passed through.

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
wmangimelli: I ended up adding the onChange back to the api, as it wasn't being recognized.

PR #127, opened 2020-08-18
Add InteractionFeedback to bundle for external use

aVileBroker:
InteractionFeedback component is useful internally for buttons and
cards, but users may also find it useful in their own components


package.json
```diff
@@ -11,8 +11,8 @@
     "build-react": "cd packages/hs-react-ui && npm run bundle",
     "prepare": "cd packages/hs-react-ui && npm install",
     "test-react": "cd packages/hs-react-ui && npm test",
-    "publish-react": "cd packages/hs-rect-ui && npm run semver-release",
-    "publish-react:dryrun": "cd packages/hs-rect-ui && npm run semver-dryrun",
```
joien1, line 15: get rect

PR #132, opened 2020-08-18
FOUNDRYUI-40 Add jest-coverage-badges module

wmangimelli:


packages/hs-react-ui/package.json
```diff
@@ -98,6 +98,7 @@
     "@types/lodash.debounce": "^4.0.6",
     "@typescript-eslint/eslint-plugin": "^3.0.0",
     "@typescript-eslint/parser": "^3.0.0",
+    "jest-coverage-badges": "^1.1.2",
```
donaldjbrady, line 101: Can we make this a dev dependency?

packages/hs-react-ui/package.json
```diff
@@ -123,7 +124,9 @@
     "webpack": "webpack --config webpack.config.js",
     "dts": "tsc --declaration --outDir dist/types/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false",
     "pre-push": "yarn run lint-ci && yarn run test && yarn run clean && yarn run webpack && yarn run dts",
-    "test": "rm -rf ./coverage/* && jest --collectCoverage=true",
+    "test": "rm -rf ./coverage/* && jest",
```
donaldjbrady, line 127: `Test` should run both `test:coverage` and `test:badges` for our build process. The `test` script gets called during our CICD pipeline

packages/hs-react-ui/package.json
```diff
@@ -123,7 +124,9 @@
...
+    "test:coverage": "npm test -- --coverage",
```
donaldjbrady, line 128: I think you have an extra `--` here

packages/hs-react-ui/package.json
wmangimelli: Thanks for catching this.

packages/hs-react-ui/package.json
wmangimelli: I will append `&& jest-coverage-badges` to the **_test_** script so the badges are updated in the CI pipeline.

packages/hs-react-ui/package.json
wmangimelli: Removed

PR #133, opened 2020-08-18
FOUNDRYUI-80 add markers to range slider

donaldjbrady:
feat(rangeslider): add markers API

### DIFF
--- a/packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
+++ b/packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
@@ -20,6 +20,7 @@ import {
   DomainLabelProps,
 } from './types';
 import { useTheme } from '../../context';
+import { Div } from '../../htmlElements';
 
 export const Container = styled.div`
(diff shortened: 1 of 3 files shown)


packages/hs-react-ui/src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -156,13 +159,43 @@ export const DomainLabel = styled.div`
...
+
+export const RangeSlider = ({
   StyledContainer = Container,
   StyledDragHandle = DragHandle,
   StyledHandleLabel = HandleLabel,
   StyledSlideRail = SlideRail,
   StyledSelectedRangeRail = SelectedRangeRail,
   StyledDomainLabel = DomainLabel,
+  StyledMarker = Marker,
+  StyledMarkerLabel = MarkerLabel,
```
joien1, line 198: These should get markerProps and markerLabelProps as well

donaldjbrady: great catch!

PR #134, opened 2020-08-18
FOUNDRYUI-56 remove table row height 100% -> master

donaldjbrady:
fix(table): remove table row height 100%


packages/hs-react-ui/src/components/Table/Table.tsx
```diff
@@ -155,12 +155,12 @@ export const Row = styled(TR)`
       column-gap: ${columnGap};
       position: relative;
       background-color: ${background};
-      height: ${isCollapsed ? '0px' : '100%'};
-  
+      height: ${isCollapsed ? '0px' : undefined};
+
```
wmangimelli, line 159: Is there a way to write this like: 
```suggestion
${isCollapsed && 'height: 0px;'}
```
so we don't assign an attribute to undefined?

PR #142, opened 2020-08-19
WIP Documentation for v1 release

wmangimelli:


packages/hs-react-ui/package.json
```diff
@@ -125,7 +126,7 @@
     "webpack": "webpack --config webpack.config.js",
     "dts": "tsc --declaration --outDir dist/types/ --emitDeclarationOnly --declarationMap --allowJs false --checkJs false",
     "pre-push": "yarn run lint-ci && yarn run test && yarn run clean && yarn run webpack && yarn run dts",
-    "test": "rm -rf ./coverage/* && jest --collectCoverage=true",
+    "test": "rm -rf ./coverage/* && jest && jest-coverage-badges",
```
donaldjbrady, line 129: will this still collect coverage like `jest --collectCoverage=true`?

wmangimelli: No, it will only collect what is specified in the jest.config, which is only set with 'json-summary' which is the report type that jest-coverage-badges uses.

PR #149, opened 2020-08-20
FOUNDRYUI-75 card interaction feedback

donaldjbrady:


packages/hs-react-ui/src/components/Card/Card.tsx
```diff
@@ -20,6 +33,19 @@ export const CardContainer = styled(Div)`
...
+      &:active {
+        background-color: ${
+          colors.background !== 'transparent'
+            ? darken(0.1, colors.background)
+            : 'rgba(0, 0, 0, 0.1)'
+        };
+      }
+    `
+          : ''
+      }
```
joien1, line 48: Does this mean that when the card is clicked that it'll darken? Is that a behavior that we want? I know this came from the button's CSS but I'm thinking that kind of behavior would look a little odd on the card.

donaldjbrady: yes, it does. The ticket was to provide interaction feedback when the card is clicked and only when an onClick function is provided by the end user (so the default onClick doesn't cause it). I saw what you did with the button's interaction feedback making it simple or ripple and did the same here to be consistent. The card defaults to ripple. There is a prop to disable the feedback entirely as well `disableFeedback` that is just a boolean. I'm happy to talk about what we expect!

PR #165, opened 2020-08-31
70 -> v1-release Card header padding

wmangimelli:
Screen Shot 2020-08-31 at 5 51 14 PM
Screen Shot 2020-08-31 at 5 52 29 PM
Screen Shot 2020-08-31 at 5 51 56 PM


packages/hs-react-ui/src/components/Card/Card.tsx
```diff
@@ -144,6 +144,10 @@ const Card = ({
   disableFeedback,
   feedbackType = FeedbackTypes.ripple,
 }: CardProps): JSX.Element | null => {
+  const hasContent = Boolean(children);
+  const hasFooter = Boolean(footer);
+  // eslint-disable-next-line no-param-reassign
+  headerProps = { ...headerProps, hasContent, hasFooter };
```
aVileBroker, line 150: I'd just declare hasContent and hasFooter above the spread, and include them on the component declaration (Line 194) - and also pass both of those variables to the footer/body as well for people to utilize when writing their own styled components

packages/hs-react-ui/src/components/Card/Card.tsx
```diff
@@ -144,6 +144,10 @@ const Card = ({
   disableFeedback,
   feedbackType = FeedbackTypes.ripple,
 }: CardProps): JSX.Element | null => {
+  const hasContent = Boolean(children);
```
aVileBroker, line 147: would either call this `hasBody` or `hasChildren` to be specific to the nomenclature here 😄

wmangimelli: Good call.

packages/hs-react-ui/src/components/Card/Card.tsx
wmangimelli: On it!

wmangimelli: @aVileBroker And should I go ahead and make a hasHeader to pass the Body and Footer?

packages/hs-react-ui/src/components/Card/Card.tsx
wmangimelli: I will name it hasBody

packages/hs-react-ui/src/components/Card/Card.tsx
aVileBroker: Yeah I think that's a good idea

packages/hs-react-ui/src/components/Card/Card.tsx
```diff
@@ -51,11 +51,11 @@ export const CardContainer = styled(Div)`
 `;
 
 export const Header = styled(Div)`
-  ${() => {
+  ${({ hasBody, hasFooter }) => {
     const { colors } = useTheme();
 
     return `
-      padding: 1.5rem 1.5rem 0rem;
+      padding: 1.5rem 1.5rem ${!hasBody && !hasFooter ? '1.5rem' : '0rem'};
```
aVileBroker, line 58: Could you add some snapshots to cover the cases where it has a body or a footer when there's a header? Maybe some of the other permutations

wmangimelli: I add the some snapshot tests around this case. Also, I updated this statement to make it more clear.

aVileBroker (approved):

PR #170, opened 2020-09-13
148 -> v1-release  Text Input uncontrolled

wmangimelli:
- [x] Add nativeAttribute.defaultValue
- [x] Rename displayValue to something more fitting, e.g. internalValue


aVileBroker:
I think this solution makes it look like an uncontrolled input but it's still controlled. I would think that we should be trying to pass no value to the native input if none was passed into the component - so the browser can do its usual text input stuff and we just report the onChange as usual.

Keeping the defaultValue is important for having uncontrolled inputs though

aVileBroker:
After talking it through - this is actually really close. Just need:

- [ ] rename displayName/setDisplayName as we're no longer displaying it
- [ ] instead of `nativeHTMLAttributes.value || ''`, do `nativeHTMLAttributes.value || nativeHTMLAttributes.defaultValue`

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -186,13 +187,12 @@ const TextInput = ({
   } else if (isMultiline) {
     InputComponent = TextAreaInputContainer;
   }
-  const displayValue = nativeHTMLAttributes.value || defaultValue;
+  const [displayValue, setDisplayValue] = useState(nativeHTMLAttributes.value || '');
```
aVileBroker, line 190:
```suggestion
  const [internalValue, setInternalValue] = useState(nativeHTMLAttributes.value || nativeHTMLAttributes.defaultValue);
```

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -204,8 +204,9 @@ const TextInput = ({
               ? e.target.value
               : e.target.value.slice(0, maxLength);
           }
-          if (nativeHTMLAttributes.onChange) {
-            nativeHTMLAttributes.onChange(e);
+          setDisplayValue(e.target.value);
```
aVileBroker, line 207:
```suggestion
          setInternalValue(e.target.value);
```

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -144,14 +145,14 @@ const defaultCallback = () => {}; // eslint-disable-line @typescript-eslint/no-e
...
   // Destructure native HTML attributes to provide default values
-  defaultValue = '',
   type = 'text',
   disabled = false,
   cols = 10,
   rows = 10,
 
   debouncedOnChange = defaultCallback,
   onClear,
+  onChange,
```
aVileBroker, line 155: what's the advantage of pulling this out of the nativeHTMLAttributes?

aVileBroker (changes requested):

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
wmangimelli: I had this within before and things went awry. When it was NOT destructed out, customizing the onChange on InputComponent would always get overwritten when spreading the native props at the _end_ of the listed props. We could keep it in the nativeAttributes, ie not destruct it out, and make sure we add our customized props for the Input Component, including onChange, after the spread of the props so the custom props take precedence. I will remove the destructed onChange, and instead move the {...nativeAtrributes} to the first listed prop on InputComponent

aVileBroker: I think your way works just fine

packages/hs-react-ui/src/components/TextInput/TextInput.tsx
```diff
@@ -179,35 +173,40 @@ const TextInput = ({
...
       {iconPrefix && createIcon(StyledIconContainer, iconPrefix)}
       <InputComponent
-        value={displayValue}
-        type={type}
-        disabled={disabled}
-        cols={cols}
-        rows={rows}
+        // Set default values above nativeHTMLAttributes
+        type="text"
+        disable={false}
```
aVileBroker, line 191:
```suggestion
        disabled={false}
```

aVileBroker (changes requested):

aVileBroker (approved):

PR #175, opened 2020-09-15
149 -> Dropdown StyledPlacedholder

wmangimelli:
UPDATED 9/22 Demonstrate that StyledPlaceholder can be passed with Dropdown.Placeholder and fix onclear button that wasn't receiving clicks.
Screen Shot 2020-09-22 at 12 03 31 PM


packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -72,12 +73,16 @@ const teaOptions = [
   },
 ];
 
+const StyledPlaceHolder = styled.span`
+  color: blue;
+`;
```
aVileBroker, line 78: while this does demonstrate the capability, I don't think this looks particularly good and in general I think we should be keeping the storybook closer to the defaults, and add in better documentation elsewhere to describe what can be customized

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -72,12 +73,16 @@ const teaOptions = [
...
+
```
aVileBroker, line 79:
```suggestion
```

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -98,6 +103,7 @@ storiesOf('Dropdown', module)
               optionsVariant={select('optionsVariant', variants, variants.outline)}
               valueVariant={select('valueVariant', variants, variants.text)}
               values={values}
+              StyledPlaceholder={StyledPlaceHolder}
```
aVileBroker, line 106:
```suggestion
```

aVileBroker (changes requested):

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -72,12 +73,16 @@ const teaOptions = [
   },
 ];
 
+const StyledPlaceholder = styled(Dropdown.Placeholder)`
+  color: ${colors.primary};
+`;
+
```
aVileBroker, line 79: I vote we don't have a fancy placeholder for the storybook at all - it's hard to know from looking at the component what is/isn't default otherwise
```suggestion
```

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -98,6 +103,7 @@ storiesOf('Dropdown', module)
               optionsVariant={select('optionsVariant', variants, variants.outline)}
               valueVariant={select('valueVariant', variants, variants.text)}
               values={values}
+              StyledPlaceholder={StyledPlaceholder}
```
aVileBroker, line 106:
```suggestion
```

aVileBroker (changes requested):

packages/hs-react-ui/src/components/Dropdown/Dropdown.stories.tsx
wmangimelli: Sounds good to me.

aVileBroker (approved):

PR #176, opened 2020-09-21
61 -> v1-release Button Styled IconContainers

wmangimelli:


packages/hs-react-ui/src/components/Button/Button.stories.tsx
```diff
@@ -20,6 +21,15 @@ const design = {
   url: 'https://www.figma.com/file/3r2G00brulOwr9j7F6JF59/Generic-UI-Style?node-id=83%3A2',
 };
 
+const StyledIconContainer = styled.div`
+  z-index: 1;
+  margin-left: 1em;
+  transition: all 0.5s ease;
```
aVileBroker, line 27: Best to not do `all` as it has performance problems. You can specify here transform since it's the only property being animated.
```suggestion
  transition: transform 0.5s ease;
```

aVileBroker (changes requested):

aVileBroker (approved):

PR #203, opened 2020-10-18
#202 fix(dropdown): merge valueItemContainerProps more intelligently

### LINKED ISSUE #202: Dropdown | valueContainer props aren't applied to the Button correctly
While we could pass the valueContainerProps to the containerProps of Button - this would overwrite the isOpen data. We need to merge the props.
image

aVileBroker:
fix #202

### DIFF
--- a/packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
+++ b/packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
@@ -423,9 +423,6 @@ const Dropdown = ({
     >
       <Button
         StyledContainer={StyledValueContainer}
-        containerProps={{
-          isOpen,
-        }}
         id={`${name}-button-value`}
         color={defaultedColor}
         onClick={(e: React.MouseEvent) => e.preventDefault()}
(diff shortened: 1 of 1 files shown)


packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -423,16 +423,19 @@ const Dropdown = ({
...
         onClick={(e: React.MouseEvent) => e.preventDefault()}
         onMouseDown={clickHandler}
         variant={variant}
         containerRef={valueContainerRef}
         {...valueContainerProps}
+        containerProps={{
+          isOpen,
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-expect-error
+          ...(valueContainerProps ? valueContainerProps.containerProps : {}),
```
aVileBroker, line 437: It is worried about valueContainerProps being undefined, so I covered for that possibility and threw an expect error on it

donaldjbrady (approved):

packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
joien1: I am unsure what the error here would be. If valueContainerProps is undefined, it should choose the second option in the ternary, right?

aVileBroker: Yep - typescript isn't that smart unfortunately. Of course the "right" option would probably be to improve the typing here - but we're using a generic subcomponent type

Issue #214, opened 2021-03-01
Dropdown | should have ability to remember scroll position


aVileBroker:
Current: The dropdown options are scrolled to the top every time they are opened (because the component is unmounted and remounted)

Expected: By default the scrolled position should be remembered and restored when a dropdown is opened. The behavior should be optionally disabled by a prop

`resetScrollPositionOnOpen?: boolean = false`

PR #215, opened 2021-03-04
#214 fix: 🐛 Scroll position now persists when open/closing options

aVileBroker:
Optionally, based on rememberScrollPosition, the dropdown options box will scroll to where it was previously scrolled to. This is useful in many standard cases, but when dynamic options are being passed in, remembering the scroll position can be harmful to the UX - hence the optional prop.

✅ Closes: #214

### DIFF
--- a/packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
+++ b/packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
@@ -212,6 +212,8 @@ export interface DropdownProps {
   onClear?: () => void;
   onSelect: (selected?: Array<string | number>) => void;
 
+  rememberScrollPosition?: boolean;
+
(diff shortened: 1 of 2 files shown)


packages/foundry-react-ui/src/components/Dropdown/Dropdown.tsx
```diff
@@ -279,6 +285,10 @@ const Dropdown = ({
     optionsHash[option.id] = { ...option, isSelected: values.includes(option.id) };
   });
 
+  const scrollListener = () => {
+    scrollPos.current = optionsContainerInternalRef.current?.scrollTop || 0;
```
joien1, line 289: This is my first time seeing ?. in the wild 😍

joien1 (approved):

PR #220, opened 2021-03-22
Per ticket requirements

jjrockhill:
- Added full-screen element within story
- Improved calculations for animations
- Modified base defaults to be a little smoother


aVileBroker:
Looks great, thanks @jjrockhill! Looks like you just need to update the snapshot. You'll need to use jest directly with `jest -u` in the `foundry-react-ui` directory and it should be good to go!

aVileBroker (approved):
Looks good, will merge once the snapshot is updated

packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -36,7 +36,8 @@ export type InteractionFeedbackProps = {
 };
 
 const defaultInterpolationFunctions = {
-  r: (r: any) => r.to((val: string) => `${Math.abs(parseFloat(val))}`),
+  r: (r: any) => r.to((val: string) => `${Math.abs(parseFloat(val).toFixed(2))}`),
```
aVileBroker, line 39: I'm seeing a type error on my end, `toFixed` returns a string which isn't accepted by `Math.abs`.

packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -36,7 +36,8 @@ export type InteractionFeedbackProps = {
...
+  opacity: (opacity: any) => opacity.to((val: string) => `${parseFloat(val).toFixed(2)}`),
```
aVileBroker, line 40: Similarly, we probably don't need the opacity string to be wrapped in a template string
 ```suggestion
  opacity: (opacity: any) => opacity.to((val: string) => parseFloat(val).toFixed(2),
```

**Review by aVileBroker — CHANGES_REQUESTED** [review]
Noticed a type problem when I ran it locally, should be fairly simple to fix.

File: packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx (hunk 1 above)
Comment (jjrockhill) [reply packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx]: Nice catch. I'm a rubyist so it would have either worked by some sort of ✨ Rails Magic ✨, resulted in a rounding error that only happens 1-out-of-10000 calculations or would have completely taken down the entire application. ;)

File: packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx (hunk 2 above)
Comment (jjrockhill) [reply packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx]: Thanks for the patience. I have not used React since 2015 and even then I think I made an increment counter.

File: packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx (hunk 1 above)
Comment (aVileBroker) [reply packages/foundry-react-ui/src/components/InteractionFeedback/InteractionFeedback.tsx]: Haha yeah and JavaScript will absolutely figure this out, but TypeScript just keeping us honest and not relying on JS magic too much.

**Comment by aVileBroker** [comment]
@jjrockhill I looked at the types a bit more locally where I can debug in VSCode, looks like this makes it happy without changing the logic:

```js
  r: (r: any) => r.to((val: string) => `${Math.abs(parseFloat(val)).toFixed(1)}`),
  opacity: (opacity: any) => opacity.to((val: number) => val.toFixed(2)),
```

aVileBroker (approved):

aVileBroker:
> Ope, we got a green check mark... GO GO GO!

Thanks for the PR @jjrockhill!

PR #235, opened 2021-06-29
#233 Set layout containment for Card component

### LINKED ISSUE #233: Card | Interaction feedback splash isn't contained to card when elevation = 0
It's not just the visual component either - the clickable area extends to the ENTIRE screen capturing all clicks. This isn't quite an edge case, but WILL break sites which use elevation 0.

Elevation 0
image

Default elevation
image

dawsonbooth:
Fixes #233 by adding the `contain: layout;` style to the Card component's top-level container.

Since the `StyledFeedbackContainer` has an absolute position, the clickable element for the ripple interaction fills the screen when the `CardContainer` component is lacking a layout restriction. When the `elevation` prop > 0, the `box-shadow` on the Card container is replaced with a `filter: drop-shadow`, which provides this restriction. Adding a layout containment to the `CardContainer` component restricts this positioning with intention and does so for all elevation values.

Some helpful info:
- [CSS Positioning](https://www.w3schools.com/css/css_positioning.asp)
- [Why does applying a CSS-Filter on the parent break the child positioning?](https://stackoverflow.com/a/52937920)

### DIFF
--- a/src/components/Button/__tests__/__snapshots__/Button.test.tsx.snap
+++ b/src/components/Button/__tests__/__snapshots__/Button.test.tsx.snap
@@ -34,6 +34,7 @@ exports[`Button keeps the container the same when switching between isLoading an
   border-radius: 0.25em;
   -webkit-transition: background-color .2s, color .5s, outline .5s, filter .5s, box-shadow .5s;
   transition: background-color .2s, color .5s, outline .5s, filter .5s, box-shadow .5s;
+  contain: layout;
   outline: 0 none;
   border: 0 none;
   cursor: pointer;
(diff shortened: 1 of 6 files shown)


aVileBroker:
This context is dope, thanks Dawson! You're not the first Headstormer to teach me something about CSS but you're one of an elite few.

aVileBroker:
Is it worth putting this in the shadow styles function so that everything that conditionally has drop-shadow/box-shadow will "just work"? I think Button might also suffer from this problem if this is the cause.

aVileBroker (changes requested):
Move the `contain: layout` style to the shadow function

dawsonbooth:
Sounds good. Any reason a drop-shadow is used instead of box shadow for elevation > 0? With the layout containment, a box-shadow without `inset` will work just as well concerning the positioning.

aVileBroker:
> Any reason a drop-shadow is used instead of box shadow for elevation > 0?

I just like drop-shadow more for more flexible shape options. It's part of the "shadows just work" thing.
You can read more on this medium article I wrote a while back: https://medium.com/headstorm/shadows-that-make-your-designers-happy-css-svg-974334186b0a

aVileBroker (approved):

PR #236, opened 2021-07-01
Update model.tsx

ifarazkhalid:
added the functionality for escape key, onClickOutside called when esc is pressed. Works now.


aVileBroker (approved):
I thought it was interesting that you used `document.addEvent...` instead of `window.addEvent...` - was there a reason you went with it? There isn't much of a difference though I usually use window out of habit. Apparently there _are_ differences, but mainly differences in the subset of events that can be called

https://stackoverflow.com/a/12045503

PR #237, opened 2021-07-01
Change child of Text component to be ReactNode

twelvetosix:


src/components/Text/Text.tsx
```diff
@@ -16,7 +16,7 @@ export const TextContainer = styled(Span)`
 `;
 
 export interface TextProps {
-  children?: string | Node;
+  children?: React.ReactNode;
```
aVileBroker, line 19: Could you import ReactNode from the React import at the top of the file?

it'd be like `import React, { ReactNode, whateverOtherReactImports } from 'react';`

aVileBroker (changes requested):
smol request :)

src/components/Text/Text.tsx
twelvetosix: This is done. I also just made it just `ReactNode` instead of `React.ReactNode`. Do you want me to go through the other files so this "style" is used throughout all of the code?

aVileBroker: Sure that'd be great! I'll approve your PR either way

aVileBroker (approved):

PR #241, opened 2021-07-12
210 dropdown refactor focus

etallion:
Prior behavior: tabbing would focus the dropdown and open it, subsequent tabbing would focus on each
item of the list. Instead, the desired behavior is one tab opens the list, only the up and down keys
can be used to navigate the list, then the next tab would close list.

 For this refactor I set the tabIndex to -1 to disable focus on the list items. I also removed the tabIndex from the Container
and adjusted the keyHandler to assign focus from the button instead of the container. This was
because there was always an extra tab needed to close and leave the dropdown because both the
container and button were being traversed. Since the button needs the focus--disabling wasn't an
option.


aVileBroker:
* Move the onFocus and onBlur off of the StyledContainer to the button
* Add an onFocus prop to the Dropdown, include in the proptypes etc like onBlur, call it if it exists while handling the button focus.

aVileBroker (changes requested):

src/components/Button/__tests__/Button.test.tsx
```diff
@@ -61,7 +73,7 @@ describe('Button', () => {
 
   it('shows ButtonContainer ButtonVariant.outline', async () => {
     const { container, getByTestId } = render(
-      <Button variant={variants.outline} onClick={() => {}} />,
+      <Button dataTestId={testId} variant={variants.outline} onClick={() => {}} />,
```
joien1, line 76: This is not a comment for changes you should make (just a mini-rant), but there's a trend in testing in libraries like this (and what this testing library was designed for) where you pretend like you're doing an integration test and do stuff like "Find the text that says "Click me" and click on that element and make sure the thing works as intended. 

I'd love to see us using data-test-ids as little as possible, since our users aren't clicking on elements with a particular test-id. I know this way of doing things was already there so I'm not expecting or asking for a change but just wanted to call that out

cc @aVileBroker

src/components/Dropdown/Dropdown.tsx
```diff
@@ -252,9 +265,11 @@ const Dropdown = ({
   color,
   elevation = 0,
   multi = false,
-  name,
+  name = 'dropdown',
```
joien1, line 268: Is this something we should generate a unique value for? What are we getting by defaulting this here? 

and if the dev does not specify the name attribute do we run into unintended issues if there are multiple dropdowns  with the same name on the page?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -300,17 +316,29 @@ const Dropdown = ({
     (e: React.FocusEvent) => {
       e.preventDefault();
       const target = e.nativeEvent.relatedTarget as HTMLElement | null;
+
       // check if we're focusing on something we don't control
-      if (!target || (target.id && !target.id.startsWith(name))) {
+      if (
+        !target ||
+        !target.getAttribute('data-test-id') ||
+        !(target.getAttribute('data-test-id') || '').startsWith(internalId.current)
```
joien1, line 324: Does this work as intended if we focus on another foundry element that also has a data-test-id?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -339,20 +367,12 @@ const Dropdown = ({
...
-      }
-    },
-    [containerInternalRef, isOpen, setIsOpen],
-  );
+  const clickHandler = useCallback(() => {
+    if (containerInternalRef && containerInternalRef.current) {
+      // Focus the container even when clicking
+      containerInternalRef.current.focus();
+    }
+  }, [containerInternalRef]);
```
joien1, line 375: IIRC refs changing don't trigger re-renders (but it's been a moment so @aVileBroker tell me if I'm wrong) so I _think_ L375 can be an empty array

src/components/Dropdown/Dropdown.tsx
```diff
@@ -481,23 +496,30 @@ const Dropdown = ({
...
                   variant={valueVariant}
+                  dataTestId={`${internalId.current}-tag`}
                   {...valueItemTagProps}
                   containerProps={{
-                    ...tagContainerItemProps,
                     dropdownVariant: variant,
                     tagVariant: valueVariant,
                     dropdownColor: defaultedColor,
                     transparentColor: colors.transparent,
+                    ...tagContainerItemProps,
```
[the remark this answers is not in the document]
joien1: This makes `dropdownVariant` etc. get overwritten by the itemProps, this seems like the intended behavior to me? So if they want to specify the itemProps variant they can do that now

src/components/Dropdown/Dropdown.tsx
```diff
@@ -518,10 +541,10 @@ const Dropdown = ({
           {options.map(option => (
             <StyledOptionItem
               id={`${name}-option-${option.id}`}
+              data-test-id={`${internalId.current}-option-${option.id}`}
               key={`${name}-option-${option.id}`}
-              onBlur={handleBlur}
               onClick={() => handleSelect(option.id)}
-              tabIndex={tabIndex}
+              tabIndex={-1}
```
joien1, line 547: Just want to make sure this still works with tabbing to the dropdown after this change, (can you confirm that for me or send me to a place I can test for myself)?

src/components/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -148,34 +153,39 @@ describe('Dropdown', () => {
...
+        onSelect={mockedSelectHandler}
+        multi
+        dataTestId="choosePokemon"
+        options={pokeOptions}
+      />,
     );
 
     act(() => {
-      fireEvent.focus(getByTestId('choosePokemon-container'));
+      fireEvent.focus(getByTestId('choosePokemon-dropdown-button'));
```
joien1, line 165: This test-id change ends up being a breaking change, is there a reason we need to make this change? I'd rather keep defaults the same wherever possible

joien1 (changes requested):
Requesting changes just to make sure questions get answered, @aVileBroker weigh in where you deem necessary

src/components/Button/__tests__/Button.test.tsx
aVileBroker: Yeah I'm using dataTestId here more so that we can check if we're clicking blurring to another element which isn't part of _this_ dropdown. I don't like using name, because it means it would need to be a required field or be a nanoid random defualt - but name attributes do actually mean things in forms so I don't like just having it be randomized. This is also why I made name be optional and default to "dropdown.

I'm fine with using a different attribute than dataTestId though. (I'm not even using dataTestId in the tests lol)

src/components/Dropdown/Dropdown.tsx
aVileBroker: https://github.com/Headstorm/foundry-ui/pull/241#discussion_r678444878

src/components/Dropdown/Dropdown.tsx
aVileBroker: Yes, because the internalId is either from a nanoid or defined by the user (and expected to be unique)

If the user passes two dropdowns the same dataTestId prop then it wouldn't work as intended.

Maybe it's worth not letting the user customize the focusing id string

src/components/Dropdown/Dropdown.tsx
aVileBroker: You can tab TO the dropdown, which opens it. Then the arrow keys control the options list, and Enter selects/deselects the options. Pressing Tab again closes the dropdown and moves to the next input/tabIndex on the page.

aVileBroker: This was the original problem the PR was made to fix. Before, tab wouldn't close the dropdown, and you'd have to go through all the options using tab before you could leave.

src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
aVileBroker: Yeah I could go either way. I made the change and then commented that I didn't like it - but yeah I guess we can leave it as is.

src/components/Dropdown/__tests__/Dropdown.test.tsx
aVileBroker: It's not that the test-id changed, we're just not focusing the container anymore.

aVileBroker (approved):

Issue #250, opened 2021-07-20
Change default font to something installed in chrome/ff/ie by default


aVileBroker:
Currently, the globalStyles default installs Monserrat from the google fonts foundry - and it adds the @import line on every component.

I'm not sure how impactful that is but I know for sure it increases the load time for pages that use foundry-ui, regardless of what global font they override. So even if they never use Montserrat, we load it on every page that has a foundry component on it.

Instead of trying to be fancy and removing the @import if the devs replace Montserrat as the default, we should just remove it completely and have a sensible default that uses the user's local fonts and has enough fallback to not be an issue. Plus Montserrat is kinda ugly anyway so it won't be missed much.

Granted this will change all the fonts for all apps which have used foundry-ui and not overwritten the default font - but if they were okay with whatever default we had before, I can't see it being an issue. If anything this may only constitute a minor version update because they could still just add the import and choose montserrat if they were really in love with it.

Example: Here's a card component after a font family override in the globalstyles.
![image](https://user-images.githubusercontent.com/7891559/126398172-92b4d029-f3e0-43f3-96b2-9800ce8ae457.png)

PR #251, opened 2021-07-21
feat(globalstyles): reduce load times by using Arial instead of Montserrat

aVileBroker:
Previously, Montserrat was used on every foundry-ui component by default. It could be overwritten by another using our globalStyles api, but the font would still be loaded by Google Fonts. This change removes Montserrat and its loading in favor of an OS-default font: Arial. It also looks better and Headstorm doesn't use Montserrat anymore anyway.

complete #250

### DIFF
--- a/.storybook/headstormTheme.js
+++ b/.storybook/headstormTheme.js
@@ -25,7 +25,7 @@ export default create({
   appBorderRadius: 4,
 
   // Typography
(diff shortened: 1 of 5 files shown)


joien1:
> It also looks better and Headstorm doesn't use Montserrat anymore anyway.
TIL

joien1 (approved):

PR #252, opened 2021-07-22
None option for modal

j-storey5:
Sorry for all the clutter, changes got made when running yarn test and then github wanted to update every file to match the latest test date. All feature changes are in /src/components/Modal/Modal.tsx and .../Modal/Modal.stories.tsx


src/components/Modal/Modal.tsx
```diff
@@ -39,6 +39,8 @@ const CloseButtonContainer = styled(Button.Container)`
   ${({ closeButtonAttachment }: { closeButtonAttachment: string }) => {
     let distance;
     let position;
+    let display_type = 'flex'; //what is the desired default for the close button's display-type??
```
aVileBroker, line 42: default distance and position to something here
update display_type to `inline-flex`

change `display_type` to `display`

aVileBroker (changes requested):

src/components/Modal/Modal.tsx
```diff
@@ -39,7 +39,7 @@ const CloseButtonContainer = styled(Button.Container)`
   ${({ closeButtonAttachment }: { closeButtonAttachment: string }) => {
     let distance;
     let position;
-    let display_type = 'flex'; //what is the desired default for the close button's display-type??
+    let display = 'in-line flex'; //default display type
```
aVileBroker, line 42: This should be `inline-flex`, which is a display attribute
https://developer.mozilla.org/en-US/docs/Web/CSS/display
```suggestion
    let display = 'inline-flex'; //default display type
```

aVileBroker (approved):

aVileBroker (changes requested):
accidental approval - still needs inline-flex change

aVileBroker (approved):

j-storey5:
Merge Success!

PR #277, opened 2021-08-05
New Skeleton component

aVileBroker:
This should maintain backwards compatibility with the old Progress component.

Closes Progress rewrite ticket

TODO:
Figure out the best way to mark a component as deprecated (needs ticket to be added)

image


dawsonbooth:
Should the size of the skeleton depend on the content in some way? I get the following for the `Button` in the codesandbox, for example:
```typescript
import "./styles.css";

import { Button } from "@headstorm/foundry-react-ui";

export default function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
      <Button isLoading>Press me</Button>
    </div>
  );
}
```
Screen Shot 2021-08-09 at 10 52 06 AM

src/components/Button/Button.tsx
```diff
@@ -37,32 +38,41 @@ export enum ButtonTypes {
 
 export type ButtonProps = {
   StyledContainer?: string & StyledComponentBase<any, {}, ButtonContainerProps>;
-  StyledLeftIconContainer?: StyledComponentBase<any, {}>;
-  StyledRightIconContainer?: StyledComponentBase<any, {}>;
+  StyledLeftIconContainer?: StyledSubcomponentType;
+  StyledRightIconContainer?: StyledSubcomponentType;
+
+  skeletonProps?: SubcomponentPropsType;
```
dawsonbooth, line 44: Should the size of the skeleton depend on the content in some way? I get the following for the `Button` in the codesandbox, for example:
```typescript
import "./styles.css";

import { Button } from "@headstorm/foundry-react-ui";

export default function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
      <Button isLoading>Press me</Button>
    </div>
  );
}
```
Screen Shot 2021-08-09 at 10 52 06 AM
Or is it intended for the user to specify things like a max width in the `skeletonProps`?

aVileBroker: Hmm the display CSS rule on the Button's loading is supposed to be `inline-block`.. not sure when I messed that up. good catch!
image

dawsonbooth (approved):

PR #282, opened 2021-08-09
#187 Dropdown open upward

### LINKED ISSUE #187: Dropdown | open upward when space below is limited
AC:

- [ ] When near the bottom of the window, the dropdown should open upward
- [ ] Add prop to prevent this behavior if desired (`flipDirectionToStayInView` or something, default to true)

harrisonnoble:
Closes #187 

- Dropdown now has an optional prop called "shouldStayInView" (defaults to true)
- When "shouldStayInView" is set to true, the dropdown menu will automatically toggle the dropdown options to face upwards or downwards depending on the available viewport room.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -42,26 +44,30 @@ const Container = styled(Div)`
...
+            border-bottom-right-radius: 0rem;
+            border-bottom-left-radius: 0rem;
+          `
+      : `
+            border-top: 0px solid transparent;
+            border-top-right-radius: 0rem;
+            border-top-left-radius: 0rem;
+          `;
+
+    return `
```
aVileBroker, line 60:
```suggestion
  ${({ isOpen, isOpenedBelow, isHidden }) => {
    const openedDirection = isOpenedBelow ? 'bottom' : 'top';
    const openStyle = `
      border-${openedDirection}: 0px solid transparent;
      border-${openedDirection}-right-radius: 0rem;
      border-${openedDirection}-left-radius: 0rem;
    `;

    return `
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -96,9 +101,43 @@ const OptionsContainer = styled(Div)`
...
+          top: 100%;
+          border-top: 0px solid transparent;
+          border-radius: 0rem 0rem 0.25rem 0.25rem;
+        `
+        : `
+          top: auto;
+          bottom: 100%;
+          border-bottom: 0px solid transparent;
+          border-radius: 0.25rem 0.25rem 0rem 0rem;
+        `
```
aVileBroker, line 117: might be worth trying to do a similar thing as above where we conditionally switch between "bottom" and "top" here

src/components/Dropdown/Dropdown.tsx
```diff
@@ -96,9 +101,43 @@ const OptionsContainer = styled(Div)`
...
+    }
+    ${isHidden ? `visibility: hidden;` : ''}
+  `}
+`;
+
+const HiddenOptionsContainer = styled(OptionsContainer)`
+  ${({ isOpenedBelow }) => `
+    visibility: hidden;
+    height: 10rem;
```
aVileBroker, line 126:
```suggestion
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -546,47 +684,55 @@ const Dropdown = ({
...
+                    {optionsHash[option.id].isSelected && <Icon path={mdiCheck} size="1em" />}
+                  </StyledCheckContainer>
+                )}
+                <Span>{option.optionValue}</Span>
+              </StyledOptionItem>
+            ))}
+          </StyledOptionsContainer>
+          <HiddenOptionsContainer
+            ref={hiddenOptionsContainerInternalRef}
+            isOpenedBelow={!isOpenedBelow}
```
aVileBroker, line 733:
```suggestion
            isOpenedAbove={!isOpenedBelow}
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -96,9 +101,43 @@ const OptionsContainer = styled(Div)`
...
+          border-bottom: 0px solid transparent;
+          border-radius: 0.25rem 0.25rem 0rem 0rem;
+        `
+    }
+    ${isHidden ? `visibility: hidden;` : ''}
+  `}
+`;
+
+const HiddenOptionsContainer = styled(OptionsContainer)`
+  ${({ isOpenedBelow }) => `
```
aVileBroker, line 124:
```suggestion
  ${({ isOpenedAbove }) => `
```

aVileBroker:
[ ] - Should probably include the StyledHiddenOptionsContainer + props + ref to the API so it can be controlled by the consuming app if they feel like it for whatever reason

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
...
+  }, []);
+
+  const intersectionCallback = useCallback(
+    (entries: IntersectionObserverEntry[]) => {
+      if (isAware) {
+        if (entries.length === 1) {
+          const [entry] = entries;
+          // swap the dropdown to open downward if its hitting the top
+          if (
+            entry.intersectionRatio < 0.95 &&
```
aVileBroker, line 400: This would be good to expose this constant as a prop (and attach to Dropdown like `Dropdown.IntersectionThreshold`)

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
 
   const tagContainerItemProps = valueItemTagProps.containerProps || {};
 
+  // effect to determine if user is scrolling up or down
+  useEffect(() => {
+    const threshold = 0;
+    let lastScrollY = window.pageYOffset;
+    let ticking = false;
```
aVileBroker, line 359: I think `ticking` should be a ref from a useRef

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
...
+        window.requestAnimationFrame(updateScrollDir);
+        ticking = true;
+      }
+    };
+    window.addEventListener('scroll', onScroll);
+
+    return () => window.removeEventListener('scroll', onScroll);
+  }, []);
+
+  const buildThresholdArray = () => Array.from(Array(100).keys(), i => i / 100);
```
aVileBroker, line 384: Consider making 100 be a prop - like `intersectionObserverPrecision`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
...
+            entry.intersectionRatio < 0.95 &&
+            entry.target === optionsContainerInternalRef.current
+          ) {
+            if (isScrollingDown) {
+              if (!isOpenedBelow && entry.intersectionRatio < prevIntersectionRatio) {
+                setIsOpenedBelow(true);
+              }
+            }
+          }
+          setPrevIntersectionRatio(entry.intersectionRatio);
```
aVileBroker, line 409: Could this be a ref instead of state? Reasons it could need to be a state var is if the rerender is needed for the app to react to intersection changes. Reason for it to be a ref is that it probably is just used for comparisons when the intersection changes and causes a re-render anyway.

My concern is that this may cause unnecessary rerenders or double renders

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
...
+        setIsHidden(false);
+      }
+    },
+    [isOpenedBelow, isScrollingDown, prevIntersectionRatio, isAware],
+  );
+
+  const intersectObserver = useMemo(() => {
+    const observer = new IntersectionObserver(intersectionCallback, intersectOptions);
+    return observer;
+  }, [intersectOptions, intersectionCallback]);
```
aVileBroker, line 428: Prettier would make this look nicer than I have it here
```suggestion
  const intersectObserver = useMemo(() => new IntersectionObserver(intersectionCallback, intersectOptions), [intersectOptions, intersectionCallback]);
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +352,96 @@ const Dropdown = ({
...
+
+  useEffect(() => {
+    const observer = intersectObserver;
```
aVileBroker, line 431: Why rename it? Would be a bit easier to follow the code if we keep the original name.

aVileBroker (changes requested):

src/components/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -15,6 +15,27 @@ const pokeOptions = [
 
 const mockedSelectHandler = jest.fn();
 
+// Need to mock the IntersectionObserver class as it is not native to node
```
aVileBroker, line 18: We should add tests to snapshot what happens when an intersection occurs

src/components/Dropdown/__tests__/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -1279,6 +1351,9 @@ exports[`Dropdown displays all options when focused 1`] = `
         </span>
       </div>
     </div>
+    <div
```
aVileBroker, line 1354: We shouldn't render this hidden options container if the flag for being space aware is false - and add a snapshot to capture it

aVileBroker:
Looks great! Just a few small changes and a couple of tests to validate the new functionality and it'll be good to go!

harrisonnoble:
Still working on fixing those snapshots, pushed latest changes so @dawsonbooth and I can start resolving merge conflicts.

coverage/coverage-summary.json
```diff
@@ -1,31 +1,31 @@
-{"total": {"lines":{"total":802,"covered":590,"skipped":0,"pct":73.57},"statements":{"total":821,"covered":600,"skipped":0,"pct":73.08},"functions":{"total":181,"covered":129,"skipped":0,"pct":71.27},"branches":{"total":845,"covered":524,"skipped":0,"pct":62.01}}
+{"total": {"lines":{"total":809,"covered":279,"skipped":0,"pct":34.49},"statements":{"total":826,"covered":282,"skipped":0,"pct":34.14},"functions":{"total":181,"covered":54,"skipped":0,"pct":29.83},"branches":{"total":856,"covered":191,"skipped":0,"pct":22.31}}
```
aVileBroker, line 1: Did coverage drop by half? The heck happened here

coverage/coverage-summary.json
```diff
@@ -1,31 +1,31 @@
...
+,"/Users/harrisonnoble/Documents/foundry-ui/src/components/Checkbox/Checkbox.tsx": {"lines":{"total":42,"covered":0,"skipped":0,"pct":0},"functions":{"total":8,"covered":0,"skipped":0,"pct":0},"statements":{"total":44,"covered":0,"skipped":0,"pct":0},"branches":{"total":41,"covered":0,"skipped":0,"pct":0}}
+,"/Users/harrisonnoble/Documents/foundry-ui/src/components/Divider/Divider.tsx": {"lines":{"total":8,"covered":0,"skipped":0,"pct":0},"functions":{"total":2,"covered":0,"skipped":0,"pct":0},"statements":{"total":8,"covered":0,"skipped":0,"pct":0},"branches":{"total":8,"covered":0,"skipped":0,"pct":0}}
+,"/Users/harrisonnoble/Documents/foundry-ui/src/components/Dropdown/Dropdown.tsx": {"lines":{"total":188,"covered":153,"skipped":0,"pct":81.38},"functions":{"total":40,"covered":33,"skipped":0,"pct":82.5},"statements":{"total":191,"covered":156,"skipped":0,"pct":81.68},"branches":{"total":177,"covered":120,"skipped":0,"pct":67.8}}
```
aVileBroker, line 7: Dropdown went up though 💪

src/components/Dropdown/Dropdown.tsx
```diff
@@ -87,7 +82,11 @@ const ValueItem = styled(Div)`
 `;
 
 const OptionsContainer = styled(Div)`
-  ${({ color, variant, isOpenedBelow, isHidden }: UsefulDropdownState) => `
+  ${({ color, variant, isOpenedBelow, isHidden }: UsefulDropdownState) => {
+    const direction = isOpenedBelow ? 'top' : 'bottom';
+    const borderStyle = isOpenedBelow ? '0rem 0rem 0.25rem 0.25rem' : '0.25rem 0.25rem 0rem 0rem';
```
aVileBroker, line 87: `border` style usually refers to the shorthand border CSS rule, maybe we can call it border radii or something.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -102,43 +101,23 @@ const OptionsContainer = styled(Div)`
...
-        `
-        : `
-          top: auto;
-          bottom: 100%;
-          border-bottom: 0px solid transparent;
-          border-radius: 0.25rem 0.25rem 0rem 0rem;
-        `
-    }
-  `}
+    ${direction}: 100%;
```
aVileBroker, line 118: if it's opened above, shouldn't the bottom be 100%?

coverage/coverage-summary.json
harrisonnoble: Hmmm thats super weird, I just checked my local coverage summary file and coverage looks like it went up, I'll re-push!

aVileBroker (changes requested):
Lookin real good

src/components/Dropdown/Dropdown.tsx
harrisonnoble: Good point, do you think `isOpenedAbove` should be renamed to `openDirection`?

aVileBroker: I think the name is okay - or it can just isOpenedBelow to be consistent with the other dropdown - but I feel like the direction is backwards, no?

harrisonnoble: Yea it is a little backwards (just for naming, not functionality) since the hidden container is doing the opposite of what the options container does (I pass in !isOpenedBelow to this prop). I'll rename it to isOpenedBelow and add a comment where I pass in the prop saying the hidden container does the opposite of the options container.

aVileBroker: Can we just call it `isOpenedBelow` and pass it around as normal but then in the CSS just flip the ternary around

src/components/Dropdown/Dropdown.tsx
```diff
@@ -292,27 +319,33 @@ const Dropdown = ({
...
+  const ticking = useRef<boolean>(false);
   const [scrollIndex, setScrollIndex] = useState<number>(0);
 
   const [isOverflowing, setIsOverflowing] = useState<boolean>(true);
-  const [isVirtual, setIsVirtual] = useState<boolean>(virtualizeOptions);
 
-  useEffect(() => {
-    setIsVirtual(virtualizeOptions && isOverflowing);
-  }, [virtualizeOptions, isOverflowing]);
+  const isVirtual = virtualizeOptions && isOverflowing;
```
aVileBroker, line 348: everyone liked that

src/components/Dropdown/Dropdown.tsx
```diff
@@ -322,6 +355,141 @@ const Dropdown = ({
...
+  );
+
+  const intersectObserver = useMemo(
+    () => new IntersectionObserver(intersectionCallback, intersectOptions),
+    [intersectOptions, intersectionCallback],
+  );
+
+  useEffect(() => {
+    if (isOpen) {
+      const timer = window.setTimeout(() => {
```
aVileBroker, line 453:
```suggestion
      // setTimeout ensures this code renders after the initial render
      const timer = window.setTimeout(() => {
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -322,6 +355,141 @@ const Dropdown = ({
...
+
+          if (optionsContainerHeight !== hiddenContainerHeight) {
+            // height is returned in pixels
+            hiddenContainer.style.height = `${optionsContainerHeight}px`;
+          }
+        }
+
+        if (optionsContainer) {
+          intersectObserver.observe(optionsContainer);
+        }
```
aVileBroker, line 481: These two if statements can be moved into the one above

aVileBroker (changes requested):

aVileBroker (approved):

PR #284, opened 2021-08-10
Virtualize dropdown options list

### LINKED ISSUE #184: Dropdown | Virtualize options list for better performance of long lists
For long lists of options, we - and material ui - are quite inefficient at rendering them, causing long lists to be very slow/janky in opening and scrolling.

Virtualizing the options list and passing each option its current status (in view vs not in view) would be very powerful. We can roll our own intersection observer hook to achieve this, or use a package - but building our own would be the least risky approach imo.

AC:
- [ ] As an option scrolls out of view, it should be replaced with an div of the height it originally was (or another performant alternative, feel free to do some research on best practices. It would be bad if it ends up being worse on performance than doing nothing)

dawsonbooth:
Uses [react-virtuoso](https://github.com/petyosi/react-virtuoso) to set up a windowed options list container that can handle items with variable, dynamic height. This is an apparently [complicated problem](https://github.com/bvaughn/react-virtualized/issues/610#issuecomment-324890558) that required a [complicated solution](https://virtuoso.dev/blog#third-pass---rxjs-to-the-rescue), and custom logic to avoid this dependency could take a while to put together.

Closes #184


src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -11,19 +11,20 @@ import Label from '../Label';
...
+  const tempData: string[] = [];
 
   for (let i = 0; i < amount; i += 1) {
-    const item = address.city();
-    finalData.push({
-      id: item.toLowerCase(),
-      optionValue: item,
-    });
+    let item = address.city();
+    while (tempData.includes(item)) item = address.city();
```
aVileBroker, line 18: I'm surprised prettier didn't add curly brackets and break this to a separate line

src/components/Dropdown/Dropdown.tsx
```diff
@@ -411,9 +412,13 @@ const Dropdown = ({
             break;
           case 'ArrowUp':
             if (focusedElement && focusedElement.id.match(`${name}-option-.*`)) {
-              const sibling = focusedElement.previousElementSibling as HTMLElement | null;
-              if (sibling) {
-                sibling.focus();
+              const row = focusedElement.parentNode as HTMLElement | undefined;
```
aVileBroker, line 415: I hadn't even thought of the implications with keyboard controls focusing options that aren't being rendered - nicely done!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -411,9 +412,13 @@ const Dropdown = ({
...
+              const rowPrevSibling = row ? row.previousElementSibling : null;
+              if (rowPrevSibling) {
+                const toFocus = rowPrevSibling.children[0] as HTMLElement | undefined;
+                if (toFocus) {
+                  toFocus.focus();
+                }
```
aVileBroker, line 421: I hadn't even thought of the implications with keyboard controls focusing options that aren't being rendered - nicely done!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -86,7 +87,7 @@ const OptionsContainer = styled(Div)`
     position: absolute;
     top: 100%;
     left: 0px;
-    max-height: 10rem;
+    height: 10rem;
```
aVileBroker, line 90: This dramatically changes the look for Dropdowns with only 2/3 options - are we assuming that they can't change the height either?

src/components/Dropdown/Dropdown.stories.tsx
dawsonbooth: Yeah, wondered the same. Looks like I forgot to push `item` to the array anyway, so I can format it with the fix.

src/components/Dropdown/Dropdown.tsx
dawsonbooth: Hmm, didn't consider that. Specifying the height of the container [looks to be necessary](https://github.com/petyosi/react-virtuoso/issues/307#issuecomment-787873104) for the scoll/window calculations. I'll look into handling a smaller number of options that would not otherwise fill up the container.

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -117,6 +132,7 @@ export const Icons: Story = args => {
 };
 Icons.args = {
   ...Basic.args,
+  placeholder: 'Choose a rating...',
```
aVileBroker, line 135: Could this be "how much do you like tea?" or something?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -423,15 +426,20 @@ const Dropdown = ({
               // get parent before nextElementSibling because buttons are nested inside of skeletons
               const optionsContainer = button ? button.nextElementSibling : null;
               if (optionsContainer) {
-                const toFocus = optionsContainer.children[0] as HTMLElement | undefined;
+                const toFocus = optionsContainer.children[0]?.children[0]?.children[0]
+                  ?.children[0] as HTMLElement | undefined;
```
aVileBroker, line 430: Wha-what

Maybe we should have another variable for storing what this is. Like "firstOption" that describes what the great grandchild of the options container is

src/components/Dropdown/Dropdown.tsx
```diff
@@ -546,18 +562,17 @@ const Dropdown = ({
...
-          {...optionsContainerProps}
-        >
-          {options.map(option => (
+        <Virtuoso
+          data={options}
+          rangeChanged={(range: ListRange) => setScrollIndex(range.startIndex)}
+          initialTopMostItemIndex={rememberScrollPosition ? scrollIndex : 0}
+          initialItemCount={
+            typeof window !== 'undefined' && window.document && window.document.createElement
+              ? initialOptionCount
```
aVileBroker, line 571: Why do we need this? Does options.length not work in SSR/test environments? Either way it probably needs a different name if it's optional. As-is I would be confused as to if I needed to supply it or not and in which situations I should/shouldn't.

aVileBroker (changes requested):
Only other thing is if we can simulate and test the virtualization (we should be able to use the same mock class/function that Harrison is using for the flippy floppy dropdown)

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -11,33 +11,47 @@ import Label from '../Label';
...
+    let item = address.city();
+    while (tempData.includes(item)) {
+      item = address.city();
+    }
+    tempData.push(item);
   }
 
-  return finalData;
+  return tempData.map(item => ({
+    id: item.toLowerCase(),
```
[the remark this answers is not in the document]
dawsonbooth: I think this would be slower, no?

```typescript
const generateCityList = (amount: number): OptionProps[] => {
  const finalData: OptionProps[] = [];

  for (let i = 0; i < amount; i += 1) {
    let city = address.city();
    let item = {
      id: city.toLowerCase(),
      optionValue: city,
    };
    while (finalData.includes(item)) {
      city = address.city();
      item = {
        id: city.toLowerCase(),
        optionValue: city,
      };
    }
    finalData.push(item);
  }

  return finalData;
};
```

src/components/Dropdown/Dropdown.stories.tsx
dawsonbooth: Makes sense, though it could be redundant with the label text:

Screen Shot 2021-08-11 at 7 58 16 PM

src/components/Dropdown/Dropdown.tsx
dawsonbooth: [Here](https://github.com/petyosi/react-virtuoso/issues/21)'s a thread on some potential SSR issues. I suppose this logic can be removed in favor of the user being able to supply a `virtualizeOptions={false}` prop if there are any complications with the virtualization.

src/components/Dropdown/Dropdown.stories.tsx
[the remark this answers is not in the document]
aVileBroker:
```js
const generateCityList = (amount: number): OptionProps[] => {
  const tempData: OptionProps[] = [];

  for (let i = 0; i < amount; i += 1) {
    let item = address.city();
    while (tempData.includes(item)) {
      item = address.city();
    }
    tempData.push(({
      id: item.toLowerCase(),
      optionValue: item,
    });
  }
};
```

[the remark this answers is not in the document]
aVileBroker: Just moving that final map into the tempDat push

src/components/Dropdown/Dropdown.stories.tsx
aVileBroker: Fair!

src/components/Dropdown/Dropdown.tsx
aVileBroker: What would be a good value here? Just what we think will fit into the options container window initially?

src/components/Dropdown/Dropdown.tsx
aVileBroker: Short lists do look a bit odd - and I can definitely see use-cases where we want the size of the options container to expand with options instead of scroll (and not be virtualized at all)

image

src/components/Dropdown/Dropdown.tsx
```diff
@@ -349,10 +364,32 @@ const Dropdown = ({
     }
 
     setIsOpen(true);
+
+    window.setTimeout(() => {
+      const focusedElement = document.activeElement;
+
+      if (focusedElement && focusedElement.id === `${name}-dropdown-button`) {
+        const button = focusedElement.parentNode as HTMLElement | undefined;
+        const optionsContainer = button ? button.nextElementSibling : null;
```
aVileBroker, line 373: can we not use `isOpen` for the conditional and the button/optionsContainer refs for the variables here?

dawsonbooth: Hmm I'll look into it -- that's just what was in place [beforehand](https://github.com/Headstorm/foundry-ui/blob/master/src/components/Dropdown/Dropdown.tsx#L421-L424)

aVileBroker (approved):

Issue #207, opened 2021-02-17
Table | Does not support footers


aVileBroker:
User story:
> As a user of the Table component I want to be able to specify a footer through the API, separate from the `data` array.

Dev hint
> This could be done in the `column` prop (adding a `footer`property to the column object to define what string/number will be displayed there) or in a separate prop from `column` or `data` simply called `footer`. The latter would mean the column definitions could be more static and only `data` and `footer` would be likely to depend on dynamic data.

Requirements:
* Accept data for a footer row which has the same typing as the normal cell rows. It should accept new overrides for the new subcomponents.
* New `footer` data should go into the tfoot DOM element to keep it as accessible as possible.
* It should conditionally render the `tfoot` depending on if any data is being rendered in it.
* It should adhere to the responsiveness that exists in the rest of the `Table` component.
* Tests should be updated if necessary to cover the new code additions.

Design:
The FooterRow and FooterCell can be restyled Row/Cell styled components, but should use grayXLight for the background and set the font-weight to 700.
![image](https://user-images.githubusercontent.com/7891559/108270100-93646380-7134-11eb-8330-20b9526e3fe1.png)

This shouldn't be a breaking change.

j-storey5:
Started 8/5/21

PR #285, opened 2021-08-10
#207 table footer

j-storey5:
added footer support, pre 1.6 release


src/components/Table/Table.tsx
```diff
@@ -617,6 +655,32 @@ const Table = ({
...
+              {Object.keys(copiedColumns).map((headerColumnKey: string, index: number) => {
+                const RenderedFooterCell = StyledFooterCell;
+                const breakpointHit =
+                  width > (copiedColumns[headerColumnKey].minTableWidth || Infinity);
+                // columns.map return
+                return (
+                  (!copiedColumns[headerColumnKey].minTableWidth || breakpointHit) && (
+                    <RenderedFooterCell key={`f${index}`}>
+                      {console.log(`f${index}`)
+                        // FOOTER CONTENT GOES HERE
```
aVileBroker, line 674: Seems like we should add footer content here

something like 
```suggestion
                        {column.footer}
```

src/components/Table/Table.tsx
```diff
@@ -201,13 +235,16 @@ const Table = ({
   expansionIconComponent,
   minWidthBreakpoint = 640,
   sortGroups = false,
+  showFooter = true,
```
aVileBroker, line 238: We can probably just replace this with a variable inside this component which checks each column to see if it has footer data

aVileBroker (changes requested):

src/components/Table/Table.tsx
```diff
@@ -70,6 +70,40 @@ export const HeaderCell = styled(TH)`
...
+    const { colors } = useTheme();
+    return `
+      display: grid;
+      grid-template-columns: ${columnWidths};
+      padding: 0rem 2rem;
+      column-gap: ${columnGap};
+      user-select: none;
+
+      background-color: ${colors.grayXlight};
+      color: black;
```
aVileBroker, line 84: I think we have a color enum for black - if header uses the CSS color like this we should update that as well

src/components/Table/Table.tsx
```diff
@@ -70,6 +70,40 @@ export const HeaderCell = styled(TH)`
...
+    transition: background-color 0.5s;
+
+    ${sortable ? '' : 'pointer-events: none;'}
+
+    &:hover {
+      background-color: rgba(255, 255, 255, 0.2);
+    }
+  `}
+`;
+
```
aVileBroker, line 106: I don't think footers care if the column is sortable right?

src/components/Table/Table.tsx
```diff
@@ -617,6 +655,27 @@ const Table = ({
         )}
       </thead>
       {createRows()}
+      {showFooter && (
+        <tfoot>
+          {width > minWidthBreakpoint && (
```
aVileBroker, line 660: this conditional could probably be merged with the showFooter one two lines up

src/components/Table/Table.tsx
```diff
@@ -617,6 +655,27 @@ const Table = ({
...
+            <StyledFooter columnGap={columnGap} columnWidths={columnWidths}>
+              {Object.keys(copiedColumns).map((headerColumnKey: string, index: number) => {
+                const RenderedFooterCell = StyledFooterCell;
+                const breakpointHit =
+                  width > (copiedColumns[headerColumnKey].minTableWidth || Infinity);
+                // columns.map return
```
aVileBroker, line 666:
```suggestion
```

src/components/Table/types.ts
```diff
@@ -18,6 +18,7 @@ export interface columnTypes {
     width?: string;
     minTableWidth?: number;
     sortable?: boolean;
+    footerContent?: string;
```
aVileBroker, line 21: This should be ReactNode so that anything (including Dropdowns) could be passed to the footer

aVileBroker (changes requested):
Leaving my previous comment open - we shouldn't need a `showFooter` flag - we can derive that based on if any columns have a `footerContent` attribute which isn't null/undefined. You could implement this with an array reduce on the columns

aVileBroker:
Looks good! Just those last two comments

aVileBroker (approved):

PR #292, opened 2021-08-18
224 Added styleConstants object to provider

PatrickDeVries:
Accessible through useTheme(), can be set like useTheme().styleConstants = {"key": "value"}. I was able to set it in button then log it in card and see the value update.


joien1 (changes requested):
> can be set like useTheme().styleConstants = {"key": "value"}

Stay away from mutating objects in general, but especially on React props, and double especially on these top-level provider objects. The way you'd use this would be something like this:

```jsx
const myStyleConstants = {
  paddingSmall: '1rem',
  paddingLarge: '2rem'
};
...
return (
  <FoundryProvider styleConstants={myStyleConstants}>
    <MyApp />
  <FoundryProvider />
);
```

The functionality looks correct. I'd like to see an example of this added to storybook, or something documenting this.

@aVileBroker can you help me understand the use case for this? Is our intention to start using some of these constants in future updates to our components? Or, what does a consumer gain from this that they don't get from having some declared constants file in their own project?

aVileBroker:
> @aVileBroker can you help me understand the use case for this? Is our intention to start using some of these constants in future updates to our components? Or, what does a consumer gain from this that they don't get from having some declared constants file in their own project?

We do plan on adding more properties here - borderRadius for instance could be something like a color scheme where the radii presets could be "subtle, intense, moderate" or something. This PR is intended to give consuming devs to use our provider to consolidate their style/theme/brand-related constants in. Too often do I have an App.tsx which has 4 different nested context providers to get all the data I need. Having a useTheme() hook which only provides a couple of items in it seems wasteful, and allowing the devs to add their own properties seems useful and a nice way to reduce code bloat.

src/context/index.tsx
```diff
@@ -18,25 +18,33 @@ type FoundryColorsType = Record<keyof typeof colorsEnum, string>;
...
-  value?: { globalStyles?: string; colors?: Partial<Record<keyof typeof colorsEnum, string>> };
+  value?: {
+    globalStyles?: string;
+    colors?: Partial<Record<keyof typeof colorsEnum, string>>;
+    styleConstants?: {};
+  };
   children: React.ReactNode;
 }) => {
-  const { globalStyles = defaultGlobalStyles, colors = colorsEnum } = value;
+  const { globalStyles = defaultGlobalStyles, colors = colorsEnum, styleConstants = {} } = value;
```
aVileBroker, line 44: I was kind of hoping for an implementation where the value object can have any attributes, kind of like this:
```suggestion
  const { globalStyles = defaultGlobalStyles, colors = colorsEnum, ...customConstants } = value;
```

It can be typed as well: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html

aVileBroker: Does that feel weird @jakeoien ?

joien1: My initial thought is it makes more sense to have them behind a specific key in the context object. 

The existing Context behaves like a configuration object, where we're effectively changing "settings" by applying a global style and colors, and having a spread over "whatever else you put in here" seems like it doesn't match the other keys hierarchically. 

So I'd recommend to separate the name out, and it's possible that we can update components later with these sorts of style constants, and we have a key on the object already available to add those in there. Just seems cleaner to me without the spread in this case

aVileBroker: Exactly what I needed to hear, thanks!

aVileBroker (approved):

src/context/index.tsx
```diff
@@ -18,25 +18,33 @@ type FoundryColorsType = Record<keyof typeof colorsEnum, string>;
 export type FoundryContextType = {
   globalStyles: string;
   colors: FoundryColorsType;
+  styleConstants: object;
```
aVileBroker, line 21: I think the type here should be more like 
```suggestion
  styleConstants?: {[key in string]: number | string};
```

I probably got that wrong but you get the idea. TS or eslint should have complained about the use of `object` here

aVileBroker (changes requested):

aVileBroker:
@PatrickDeVries The last two things this PR needs is an update to the README file somewhere in/around global styling where the provider is mentioned - and an update to the provider story in storybook

README.md
```diff
@@ -179,7 +179,9 @@ Output:
 
 See the [Form demo example](https://headstorm.github.io/foundry-ui/?path=/story/form-example--controlled-form) and [source code](./src/components/Examples/Form.stories.tsx)
 
-### Global styling
+### Global Styling / Provider
+
+FoundryProvider provides global styling utilities and the option to store constants globally by setting styleConstants.
```
aVileBroker, line 184:
```suggestion
FoundryProvider provides global styling utilities and the option to store custom constants globally through the `styleConstants` attribute.
```

aVileBroker (approved):

joien1 (approved):

Issue #185, opened 2020-10-14
Dropdown | Add searching (with callbacks)


aVileBroker:
We should have a default internal search which sorts the current option list, which can be turned off and onChange callbacks (also pass the onDebouncedChange callback up as a separate function)

I like to use fuse.basic.js for fuzzy searching lists on client-side. It weights the results and allows for mispellings, searching multiple properties in a nested object, and weights for each property in nested objects. It also returns where in the strings the matches occur so that we can highlight/bold the letters which match the search.

The API could be

- searchable
- searchFiltersOptions (needs a better name - but this would be what tells us to filter the list ourselves, and add the boldness to the matched patterns)
- onSearchChange
- onDebouncedSearchChange

PR #296, opened 2021-08-20
#185 Add searching to dropdown

dawsonbooth:


src/components/Dropdown/Dropdown.tsx
```diff
@@ -206,6 +215,12 @@ const StyledTagContainer = styled(Tag.Container)`
   `}
 `;
 
+const StyledSearchContainer = styled(Div)``;
+
+const StyledSearchInput = styled(Input)`
+  all: inherit;
```
aVileBroker, line 221: ayo?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -322,8 +345,14 @@ const Dropdown = ({
 
   shouldStayInView = true,
   intersectionThreshold = 1.0,
+  intersectionContainer = null,
   intersectionObserverPrecision = 100,
   virtualizeOptions = true,
+
+  searchable = true,
```
aVileBroker, line 352: we probably don't want to default to true here - usually filtering is only needed when the user is presented with a long list.

That said if you think the experience is basically the same we can keep it on by default - but what if someone thinks they can type their value into the search box if it's not in the list 🤔 maybe we ought to clear the search string on blur?

aVileBroker (approved):

PR #299, opened 2021-08-23
287 Global Analytics Function

PatrickDeVries:
Added global analytics functionality. Default function can be overridden by end users like the example in the globalStyles story. Will be called automatically for events on components where the event has been overridden by the analytics event handler, which should be all current events.


src/components/Card/__tests__/__snapshots__/Card.test.tsx.snap
```diff
@@ -430,6 +430,7 @@ exports[`Card shows Card with non-default elevation 1`] = `
 
 exports[`Card shows Card with ripple feedback and no onClick 1`] = `
 .c0 {
+  cursor: pointer;
```
aVileBroker, line 433: Card.tsx line 42 checks if onClick === defaultOnClick and sets the cursor to pointer - we need to update it to check if it's the same as whatever the new default is so it doesn't make the cursor a pointer for all cards

src/components/Dropdown/__tests__/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -2779,6 +2779,11 @@ exports[`Dropdown renders a value when given a matching option id through props
   pointer-events: none;
 }
 
+.c8 {
+  height: 1.125rem;
+  z-index: 1;
+}
+
```
aVileBroker: Are the changes to the snapshots expected?

src/components/Dropdown/__tests__/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -2828,6 +2833,21 @@ exports[`Dropdown renders a value when given a matching option id through props
...
+            viewBox="0 0 24 24"
+          >
+            <path
+              d="M19,6.41L17.59,5L12,10.59L6.41,5L5,6.41L10.59,12L5,17.59L6.41,19L12,13.41L17.59,19L19,17.59L13.41,12L19,6.41Z"
+              style="fill: currentColor;"
+            />
+          </svg>
+        </div>
+        <div
+          class=" c8"
```
aVileBroker, line 2850: Also probably not expected

src/components/TextInput/TextInput.tsx
```diff
@@ -199,6 +199,9 @@ const TextInput = ({
   const [internalValue, setInternalValue] = useState(
     nativeHTMLAttributes.value || nativeHTMLAttributes.defaultValue || '',
   );
+  const handleEventWithAnalytics = useAnalytics();
+  const handleClear = (e: any) =>
+    handleEventWithAnalytics('TextInput', onClear, 'onClear', e, containerProps);
```
aVileBroker, line 204: We should add analytics to debouncedChange as well

src/context/index.tsx
```diff
@@ -86,3 +128,30 @@ export const withGlobalStyle = (Component: StyledSubcomponentType) => {
...
+    deviceInfo.innerHeight = window.innerHeight;
+    deviceInfo.innerWidth = window.innerWidth;
+
+    context.analyticsFunction(
+      componentType,
+      eventType,
+      event,
+      dateTime,
+      deviceInfo,
+      window.location.href,
```
aVileBroker, line 153: @dawsonbooth does accessing window like this make problems for SSR?

src/context/index.tsx
```diff
@@ -11,19 +12,51 @@ export const defaultGlobalStyles = `
...
+  currentURL,
+  name: 'name' in props ? props.name : 'No name provided',
+  analytics: 'analytics' in props ? props.analytics : 'No analytics object provided',
+});
+
 type FoundryColorsType = Record<keyof typeof colorsEnum, string>;
+type AnalyticsFunctionType = (
+  componentType: string,
+  eventType: string,
+  eventArgs: React.ChangeEvent<HTMLInputElement>,
```
aVileBroker, line 42: I don't think the type is always correct here

src/context/index.tsx
```diff
@@ -86,3 +128,30 @@ export const withGlobalStyle = (Component: StyledSubcomponentType) => {
...
+export const useAnalytics = () => {
+  const context = useContext(FoundryContext);
+  return (
+    componentType: string,
+    eventHandler: any,
+    eventType: string,
+    event: any,
+    props: any,
+  ): void => {
+    eventHandler(event);
```
aVileBroker, line 141: can we rename these variables a bit to get us to

```js
eventFunction(eventArgs)
```

aVileBroker (changes requested):

src/components/Card/__tests__/__snapshots__/Card.test.tsx.snap
PatrickDeVries: I changed it to make the comparison in the Card component and passed the result as a prop to the container

src/components/Dropdown/__tests__/__snapshots__/Dropdown.test.tsx.snap
PatrickDeVries: No they are not. Any idea why the changes would happen? I haven't touched any of the CSS to my knowledge.

src/context/index.tsx
dawsonbooth: Hmm, that looks to be the case:
- https://github.com/gatsbyjs/gatsby/issues/19294
- https://stackoverflow.com/a/66106510
SSR frameworks like Gatsby or Next will provide something like a `location` prop that can substitute for the current URL.

dawsonbooth: Next.js: https://stackoverflow.com/a/58023653
Gatsby: https://css-tricks.com/how-to-the-get-current-page-url-in-gatsby/#method-2-get-the-href-property-of-location-data-from-props

PatrickDeVries: I changed it to check if window is defined before setting the value. If undefined, the value defaults to a warning about requiring the analytics function to handle window info.

src/components/Dropdown/__tests__/__snapshots__/Dropdown.test.tsx.snap
aVileBroker: No, but adding an onClick where it is expected to be null could cause some conditional CSS to get inserted.

src/components/Card/Card.tsx
```diff
@@ -176,11 +176,17 @@ const Card = ({
   const hasBody = Boolean(children);
   const hasFooter = Boolean(footer);
 
+  const handleEventWithAnalytics = useAnalytics();
+  const handleClick = (e: any) =>
+    handleEventWithAnalytics('Card', onClick, 'onClick', e, containerProps || { name: 'Card' });
+  const isDefaultOnClick = onClick === defaultOnClick;
```
aVileBroker, line 182: this is pretty simple logic - could we move it into the StyledContainer styled-component to reduce the API footprint? I know we were doing that before, so I'm not sure why this fixed it as opposed to what was there before

src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -286,6 +286,8 @@ export const RangeSlider = ({
 
   const domain = max - min;
 
+  const handleEventWithAnalytics = useAnalytics();
+
   // set the drag value asynchronously at a lower frequency for better performance
   const valueBuffer = useRef(0);
   const debouncedDrag = debounce(() => onDrag(valueBuffer.current), debounceInterval);
```
aVileBroker, line 293: This would be good to add analytics too as well - if you want to just get the base events in and make a PR, feel free to create a new issue with any events you didn't get to for someone else to follow up with.

aVileBroker (changes requested):

src/context/index.tsx
```diff
@@ -55,12 +94,15 @@ export const FoundryProvider = ({
     ...colorsEnum,
     ...colors,
   };
+  const mergedAnalytics =
+    defaultAnalyticsFunction === analyticsFunction ? defaultAnalyticsFunction : analyticsFunction;
```
aVileBroker, line 98: is this backwards?

dawsonbooth: This reads the same as `const mergedAnalytics = analyticsFunction`, no?

PatrickDeVries: I removed it entirely and passed analyticsFunction directly. Not sure what I was doing here.

src/components/RangeSlider/RangeSlider.tsx
PatrickDeVries: I added analytics to these events I had skipped over: image
Let me know if you see any others I missed

Issue #295, opened 2021-08-19
TextInput | Bug fixes for clearing functionality


harrisonnoble:
The character count does not reset when the clear button is pressed.

![Screen Shot 2021-08-19 at 5 21 32 PM](https://user-images.githubusercontent.com/35738896/130151879-cb5434ef-9f42-4211-8521-a52eb9cf4c28.png)

![Screen Shot 2021-08-19 at 5 21 48 PM](https://user-images.githubusercontent.com/35738896/130151922-131bb387-4569-449d-87df-f94ccefe83a7.png)

This seems to be an issue with the `internalValue` state not being set when cleared.

AC:

- [x] When the TextInput is cleared, the character count should display 0
- [x] When the TextInput is cleared, the `internalValue` state should reflect the cleared text value
- [x] When using an uncontrolled TextInput (where value is undefined), having clearable === true should cause the clear X to appear and onClick it should set the input to null/undefined/''
- [x] When showCharacterCount === true, we should show the current char count part of the X/Y message even if there is no maxLength provided.
- [ ] When the clear button (and probably icons) are present, the focus highlighting is on the input only - instead of the entire container. Dev hint: Search for other components which use focus-within for implementation.

PR #305, opened 2021-08-25
fix(component): #295 set internal value of TextInput on clear

blakekrammes:
The empty value being passed from the parent on clear was not being applied since that value was
only used as an initial value for useState in the child


src/components/TextInput/TextInput.tsx
```diff
@@ -205,6 +207,8 @@ const TextInput = ({
     debounceInterval,
   ]);
 
+  const internalInputRef = useRef(inputRef?.current);
```
blakekrammes, line 210: Not sure this is the right approach or that my understanding of refs is 100%. I deleted the test checking for inputRef.current because this change causes that to fail.

src/components/TextInput/TextInput.tsx
```diff
@@ -243,23 +251,35 @@ const TextInput = ({
...
+        <StyledIconContainer
+          onClick={
+            onClear
+              ? (e: SyntheticEvent<Element, Event>) => handleClear(e)
+              : () => {
+                  if (internalInputRef?.current?.value) {
+                    setInternalValue('');
+                    internalInputRef.current.value = '';
+                  }
+                }
```
aVileBroker, line 267: I think it might be more clear for someone reading the code if this logic was all handled in the `handleClear` function

src/components/TextInput/TextInput.tsx
```diff
@@ -243,23 +251,35 @@ const TextInput = ({
...
       />
-      {onClear && (
-        <StyledIconContainer onClick={handleClear} {...iconContainerProps}>
+      {(onClear || clearable) && (
+        <StyledIconContainer
+          onClick={
+            onClear
+              ? (e: SyntheticEvent<Element, Event>) => handleClear(e)
+              : () => {
+                  if (internalInputRef?.current?.value) {
```
aVileBroker, line 263: We also want to check if value was provided through the props (i.e. controlled component) because we only want to clear the input if it's uncontrolled and value is undefined

src/components/TextInput/TextInput.tsx
```diff
@@ -243,23 +251,35 @@ const TextInput = ({
...
         <StyledCharacterCount
           ref={characterCountRef}
           errorMessage={errorMessage}
           isValid={isValid}
-          textIsTooLong={(internalValue as string).length > maxLength}
+          textIsTooLong={maxLength ? (internalValue as string).length > maxLength : false}
           {...characterCountProps}
         >
-          {(internalValue as string).length} / {maxLength}
+          {(internalValue as string).length} {maxLength ? `/ ${maxLength}` : null}
```
aVileBroker, line 282: Just since maxLength is a number and could be 0 (not sure why a textinput would have 0 as a max length but a user might)
```suggestion
          {(internalValue as string).length} {maxLength !== undefined ? `/ ${maxLength}` : null}
```

src/components/TextInput/TextInput.tsx
```diff
@@ -243,23 +251,35 @@ const TextInput = ({
...
         </StyledIconContainer>
       )}
-      {showCharacterCount && maxLength && (
+      {showCharacterCount && (
         <StyledCharacterCount
           ref={characterCountRef}
           errorMessage={errorMessage}
           isValid={isValid}
-          textIsTooLong={(internalValue as string).length > maxLength}
+          textIsTooLong={maxLength ? (internalValue as string).length > maxLength : false}
```
aVileBroker, line 279:
```suggestion
          textIsTooLong={maxLength !== undefined ? (internalValue as string).length > maxLength : false}
```

src/components/TextInput/TextInput.tsx
aVileBroker: I think you can just add a new ref and merge the two refs (there's a utility function called mergeRefs for this, check out how it's used elsewhere)

aVileBroker (changes requested):

blakekrammes:
Thanks for the feedback @aVileBroker. I have a deadline tomorrow, so we'll see, but hopefully I can get to this tonight. Just wanted to update you.

aVileBroker:
> Thanks for the feedback @aVileBroker. I have a demo tomorrow, so we'll see, but hopefully I can get to this tonight. Just wanted to update you.

No rush @blakekrammes, appreciate the extra work on this! 👍

aVileBroker (approved):

Issue #315, opened 2021-08-31
New Component | Annotation (attachment, facehugger, metroid, or other name)


aVileBroker:
This is a component which is fixed in position, and "attaches" itself to a component by ref. It could be rendered in a portal - we could probably do a ternary to control it, like `<renderInPortal ? Portal : Fragment>` .

It behaves similarly to the dropdown currently does, in that it will "flip" when it runs out of space. Ideally this should be used in Dropdown as well to replace the current intersectionObserver logic, and in doing so also complete #255 at the same time.

### Props API:
* `preferredPlacement?: 'above' | 'right' | 'bottom' | 'left';`
* `preferredAlignment?: 'start' | 'center' | 'stretch' | 'end';`
* `renderInPortal: boolean;`
* `StyledContainer?: StyledSubcomponent;`

Given the preferred placement/alignment provided by the props - it should follow those unless it runs out of space, in which case it should "flip". I don't see the need to "flip" to `center` or `stretch`, but feel free to prove me wrong.

### Rationale
I began writing this component as part of the Spotlight component for showing an annotation around the spotlight which tries to always stay in-view. When it goes near the right of the screen, it flips to the left. When it goes towards the bottom, it flips up, etc.

This is also useful for tooltips, and I'm sure our users will find other use-cases as well. This is not a simple problem to fix - which is exactly why having a component like this would make foundry powerful and sought-after.

If there is a library out there which we like, and isn't too large, I'm open to suggestion for simplifying this and potentially negating the need for it at all.

aVileBroker:
I would love to also see a mobile friendly version of this, where the annotation goes full width (w/ gutters) and only clamps to the bottom/top of the elements.

Same could be done for tight landscape views, where the annotation is full height and clamps to the left/right sides of the target

PR #318, opened 2021-08-31
301 spotlight component

aVileBroker:


aVileBroker:
Todo:
- [ ] fix snapshots
- [ ] remove annotation positions
- [ ] add analytics for onClick
- [ ] make shape options an enum
- [ ] animate the fade-in animation

aVileBroker:
The only bug here is that it can't animate from round/box/rounded box --> circular. This is fine imo, but the spring breaks when trying to do so and not only does it not animate to the circular shape, it just straight up does nothing.

dawsonbooth:
Looks like the user gets stuck and can't move forward if the nav goes off-screen:
image
Do we want to handle this similar to Harrison's PR for the dropdown?

dawsonbooth (commented):
Looks like the user gets stuck and can't move forward if the nav goes off-screen:
image
Do we want to handle this similar to Harrison's PR for the dropdown?

dawsonbooth (commented):
Getting some masking glitches. Is there any way to prevent this? 
image
image
image

aVileBroker:
@dawsonbooth I made a separate ticket for an "Annotation" component here: #315 which should handle that imo.

Perhaps the screen should scroll to the element before highlighting it. I think that much could be within the scope of this component - but the text/buttons in the story are purely in the story. This component really just handles the blurry cutout that animates and moves to the element.

aVileBroker:
As for masking glitches - I could add a bit more padding around the cutout for where the blurs are so that they don't touch it and cause semi-blurred artifacting.

aVileBroker:
@dawsonbooth I think I'd rather have the artifacting be captured in a bug ticket and the auto-scrolling into an enhancement (since that is VERY complex to figure out which scroll container needs to be scrolled to expose the element if it's not the body scroll)

dawsonbooth (changes requested):
When the shape is set to circular, there is no longer a transition animation for the spotlight circle itself, though the text and buttons move fine.

https://user-images.githubusercontent.com/10625031/132918310-77843f05-4173-4436-9b52-f9ebe0a0cbaa.mov

aVileBroker:
> When the shape is set to circular, there is no longer a transition animation for the spotlight circle itself, though the text and buttons move fine.

Ooh good catch, I'll fix that.

aVileBroker:
@dawsonbooth The circle shape now animates between targets - but still doesn't correctly switch between circle -> non-circle shapes. Unfortunately there isn't a way to fix this and won't be part of this PR.

dawsonbooth (approved):

PR #329, opened 2021-09-10
fix(textinput): swap nullish coalescing for && to fix broken bundle

aVileBroker:


aVileBroker:
Tested this fix using the codesandbox hosted package - merging without reviews 🤙

PR #339, opened 2021-10-04
302 StepProgress component

### LINKED ISSUE #302: StepProgress component
> As a developer I want to be able to show the users progress through a wizard, form flow, or other multistep process.

I'm leaving this one intentionally vague for the person who implements it to give some creative freedom on what we can do.

Things to consider:
* We probably want click events on each step with a reference passed as an argument for which step is clicked
* Indicators for complete, current, and not completed.
* Many forms let us go back to previous steps - we should be able to differentiate the current step even if it's completed.
* Animations?

Either dots, lines, etc.
Example from Dribbble: https://dribbble.com/shots/15007566-Progress-Indicator-Dark
image

Example from Dribbble: https://dribbble.com/shots/2146282-Sign-Up-Progress-Indicators
image

Example from Dribbble: https://dribbble.com/shots/10034848-Tabbar-Indicator
image

Do your own research - I always suggest Dribbble to find fun animated inspiration. Remember we want to be enable developers to do whatever they want with our components - no Dribbble shot should be impossible to recreate. (within reason)

PatrickDeVries:
Added the StepProgress component from #302. I have not written tests before so I copied the RangeSlider tests mostly, let me know if they need anything else added.


dawsonbooth (commented):
Should they be this big by default?

image

dawsonbooth (changes requested):

dawsonbooth (dismissed):

PR #363, opened 2021-12-03
Fixes #355 Fixes #356 Fixes #357 text input improvements

### LINKED ISSUE #355: TextInput | clearable X no longer works
Adding the clearable flag will add the X button - but it does nothing when clicked (doesn't call the onClear function either)

This should have been caught by unit tests so if there isn't one already, make sure this functionality is tested so we don't get further regressions.

### LINKED ISSUE #356: TextInput | iconPrefix has click pointer but does not focus the input
Clicking on the icon provided in `iconPrefix` should focus the text input just like clicking on the input would.
The icon should also not be a pointer, which suggests a "click" action like a button instead of a "focus" or "highlight" action. Having the icon's cursor be the text cursor makes more sense when the click focuses the input.

### LINKED ISSUE #357: TextInput | Limit iconPrefix options in story to improve load time
In the storybook story for TextInput the options for iconPrefix are _all_ material community icons which makes the story take a long time to open.

Look at the Dropdown and Button for examples of limited options to use, and pick icons which make sense for a text input (contact form, card number, etc)

aVileBroker:


aVileBroker:
This is giving about a 2-3% bump on test coverage 🎉

dawsonbooth (approved):

Issue #309, opened 2021-08-27
Dropdown | exceeds maximum update depth if options is undefined


aVileBroker:
options can be an empty array and works fine - but if it's undefined (and it can be based on the type) it exceeds the maximum update depth and causes the page to crash.
```
react-dom.development.js:88 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in Dropdown (at Spotlight.stories.tsx:23)
    in div (created by Styled(styled.div))
    in Styled(styled.div) (at context/index.tsx:86)
    in Div (created by Header)
    in Header (at Card.tsx:188)
    in div (created by Styled(styled.div))
    in Styled(styled.div) (at context/index.tsx:86)
    in Div (created by CardContainer)
    in CardContainer (at Card.tsx:180)
    in Card (at Spotlight.stories.tsx:22)
    in unboundStoryFn
    in ErrorBoundary
```
![image](https://user-images.githubusercontent.com/7891559/131192770-b8419c4c-b87d-4223-b8d7-04453c0b691d.png)

AdrikaKhan:
I'll take this.

PR #369, opened 2021-12-21
Issue card #309 - Changed useEffect dependency array from options to JSON.stringify(options)

jonesdebu:
And filteredOptions to JSON.stringify(filteredOptions). Also added a test to Dropdown.test.tsx to test for an infinite loop when options is undefined.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -394,13 +394,13 @@ const Dropdown = ({
   const searchInputRef = useRef<HTMLInputElement>(null);
   const [searchCharacterCount, setSearchCharacterCount] = useState<number>(0);
   const [filteredOptions, setFilteredOptions] = useState<OptionProps[]>([]);
-  const stringifyOptions = JSON.stringify(options);
+  const stringifiedOptions = JSON.stringify(options);
 
   useEffect(() => {
     setFilteredOptions(options);
     // empty array of options makes this useEffect loop
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [stringifyOptions]);
+  }, [stringifiedOptions]);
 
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -395,10 +394,13 @@ const Dropdown = ({
   const searchInputRef = useRef<HTMLInputElement>(null);
   const [searchCharacterCount, setSearchCharacterCount] = useState<number>(0);
   const [filteredOptions, setFilteredOptions] = useState<OptionProps[]>([]);
+  const stringifyOptions = JSON.stringify(options);
```
aVileBroker, line 397: this variable name sounds like a function - `stringifiedOptions` is probably better

src/components/Dropdown/Dropdown.tsx
```diff
@@ -501,6 +503,7 @@ const Dropdown = ({
     [intersectOptions, intersectionCallback],
   );
 
+  const stringifyFilteredOptions = JSON.stringify(filteredOptions);
```
aVileBroker, line 506: same comment here

src/components/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -352,6 +352,18 @@ describe('Dropdown', () => {
...
+    it('Should pass accessibility test with default props', async () => {
+      generateIntersectionObserver([]);
+      const component = (
+        <Dropdown onSelect={() => {}} placeholder="hello" options={undefined}></Dropdown>
+      );
+      const { container } = render(component);
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
```
aVileBroker, line 365: this accessibility test probably works - but I think you can just do a simple snapshot test here.

aVileBroker (changes requested):

PR #370, opened 2021-12-22
367 Use marker index for key

breanna-eubanks:
Update markers to use the array index instead of the value.


src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -518,7 +518,7 @@ export const RangeSlider = ({
 
         return (
           <StyledMarker
-            key={`marker-${value}`}
+            key={`marker-${index}`}
             id={`marker-${value}`}
```
aVileBroker, line 522: Sorry to get back to this after such a long time - do you think its worth also swapping the id to use the index? If this is the better way to identify it, I'd think it should be used for both.

```suggestion
            id={`marker-${index}`}
```

aVileBroker (changes requested):

PR #374, opened 2022-03-31
Add color prop for checkbox icon

### LINKED ISSUE #372: Checkbox | Add color prop to allow full customization of checkbox color
AC:

- Add color prop
- Use as override for icon color (regardless of varient/checkboxType)

cnkeats:
fixes #372


src/components/Checkbox/Checkbox.tsx
```diff
@@ -87,9 +87,14 @@ export const CheckboxContainer = styled.div`
...
-  height: 1em;
-  width: 1em;
+  ${({ color }) => {
+    return `
+      overflow: visible;
+      color: ${color};
+      width: 1em;
+      height: 1em;
+    `;
+  }}
```
aVileBroker, line 97: I think this can be simplified a bit

```suggestion
  ${({ color }) => `
    overflow: visible;
    color: ${color};
    width: 1em;
    height: 1em;
  `}
```

dawsonbooth: a bit more?
```typescript
    overflow: visible;
    color: ${({ color }) => color};
    width: 1em;
    height: 1em;
```

aVileBroker: Gets a bit weird when you add types to the props if you're using a type definition - it ends up being a Pick, which gets kinda awkward. That's the main reason I like having them all imported at the top

aVileBroker:
So, I know the AC in the ticket said "regardless of the variant and checkbox color" but I don't think that makes sense when actually trying it (see screenshot from codesandbox)

image

I would think the color prop would just change the general color of the checkbox (in this case where the green is), and we should be able to override the checkbox color separately

aVileBroker:
Being able to control both would give us ultimate control

https://user-images.githubusercontent.com/7891559/162487773-bf5ba8f9-1543-4a5f-b1cd-518e0967de92.mp4

src/components/Checkbox/Checkbox.tsx
dawsonbooth: Pretty sure the styled function is generic, i.e.

```js
export const MyComp = styled.div<{tall: boolean; color: string}>`
  height: ${({ tall }) => tall ? 2 : 1}rem;
  color: ${({ color }) => color};
`;
```

dawsonbooth: or maybe it gets weird because of the `as StyledSubcomponentType` cast?

PR #376, opened 2022-04-06
Fixes #375: Adding SkeletonProps

### LINKED ISSUE #375: Text | No prop to customize Skeleton
AC:

Add props:
- StyledSkeleton
- skeletonProps
- skeletonRef

As is done in other components in the same foundry-ui architecture pattern

GriseldaZBalmaceda:


src/components/Skeleton/Skeleton.tsx
```diff
@@ -68,27 +67,31 @@ const SkeletonContainer = styled(Div)`
...
+    }
   `}
 `;
 
 export type SkeletonProps = {
   StyledContainer?: StyledSubcomponentType;
   StyledShimmer?: StyledSubcomponentType;
 
-  containerProps?: SubcomponentPropsType;
+  SkeletonProps?: SubcomponentPropsType;
```
aVileBroker, line 78: I'd keep this name the same, as it matches the StyledContainer and other components where "container" is used for the outer-most element of the component.
```suggestion
  containerProps?: SubcomponentPropsType;
```

src/components/Skeleton/Skeleton.tsx
```diff
@@ -68,27 +67,31 @@ const SkeletonContainer = styled(Div)`
...
   shimmerProps?: SubcomponentPropsType;
 
   children?: React.ReactNode;
   color?: string;
   isLoading?: boolean;
+
+  skeletonRef?: React.RefObject<HTMLElement>;
```
aVileBroker, line 85: I know it's out of scope, but while you're here could you add a `containerRef`?  That would be the last thing missing on the API for this component.

src/components/Text/Text.tsx
```diff
@@ -23,17 +23,19 @@ export interface TextProps {
   isLoading?: boolean;
   isProcessing?: boolean;
   size?: string;
-
+  StyledSkeletonContainer?: StyledSubcomponentType;
   StyledContainer?: StyledSubcomponentType;
   StyledIconContainer?: StyledSubcomponentType;
   containerProps?: SubcomponentPropsType;
   iconContainerProps?: SubcomponentPropsType;
+  SkeletonProps?: SubcomponentPropsType;
```
aVileBroker, line 31:
```suggestion
  skeletonProps?: SubcomponentPropsType;
```

src/components/Text/Text.tsx
```diff
@@ -94,11 +97,12 @@ const Text = ({
         <Icon aria-hidden="true" path={mdiLoading} size={size} spin={1} />
       </StyledIconContainer>
     )}
-    <Skeleton color={color} StyledContainer={SkeletonContainer} isLoading={isLoading}>
+    <Skeleton isLoading={isLoading} skeletonRef={skeletonRef} StyledContainer={StyledSkeletonContainer} {...SkeletonProps}>
```
aVileBroker, line 100: looks like we lost the color prop here. It should still be set here by default, but now it will be able to be overwritten by passing `color` into the skeletonProps prop

aVileBroker (changes requested):
Minor capitalization changes and a request to add a `containerRef` to the Skeleton component - otherwise looks good to go!

aVileBroker (approved):

PR #377, opened 2022-04-06
fix yarn scripts for windows

### LINKED ISSUE #358: Package.json | syntax for "test" and "push" pre-commit hooks fail on Windows machines
There was an effort made by @dawsonbooth to do this, but it is currently in a state that only works in mac/linux shells. A solution that works in both would be ideal, barring that some OS detection which runs a different script?

cnkeats:
fixes #358


aVileBroker:
Tested while doing #378 on Windows and it worked great! Just need to test on mac to confirm this doesn't break it 🎉

aVileBroker (approved):

PR #387, opened 2022-05-03
Fixes #360 Stop spotlight from animating during scroll and resize events

### LINKED ISSUE #360: Spotlight | Reactions to scrolling/window resize shouldn't smoothly animate
Currently when scrolling while the spotlight is targeting an element, the spotlight will smoothly animate to the new position - which makes it feel like it's lagging behind. Ideally it should just scroll with the page contents. The same happens when the screensize changes

Dev hint: The useMemo for the spotlight rectangle bounds can update as usual, but the useSpring config needs to be `set()` with `immediate: true` when it is triggered by a scroll.

aVileBroker:
Adding @jakeoien as I'm breaking the rules of hooks and there's probably a better way of doing this.

### DIFF
--- a/src/components/Spotlight/Spotlight.tsx
+++ b/src/components/Spotlight/Spotlight.tsx
@@ -1,9 +1,10 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo } from 'react';
 import styled from 'styled-components';
 import { animated, useSpring } from '@react-spring/web';
 import { Portal } from 'react-portal';
 
 import { SubcomponentPropsType, StyledSubcomponentType } from '../commonTypes';
+import { useStateWithPrevious } from '../../utils/hooks';
 import { useAnalytics } from '../../context';
 import { AnimatedDiv } from '../../htmlElements';
 
@@ -89,11 +90,14 @@ const Spotlight = ({
   animationSpringConfig,
 }: SpotlightProps): JSX.Element | null => {
   const handleEventWithAnalytics = useAnalytics();
-  const [windowDimensions, setWindowDimensions] = useState<{ width: number; height: number }>({
+  const [windowDimensions, prevWindowDimensions, setWindowDimensions] = useStateWithPrevious<{
+    width: number;
+    height: number;
+  }>({
     width: window.innerWidth,
     height: window.innerHeight,
   });
-  const [scrollTop, setScrollTop] = useState<number>(0);
+  const [scrollTop, prevScrollTop, setScrollTop] = useStateWithPrevious<number>(0);
 
   const rect = useMemo<Pick<DOMRect, 'x' | 'y' | 'width' | 'height' | 'bottom' | 'right'>>(() => {
     const defaultVal = {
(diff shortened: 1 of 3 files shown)


src/components/Spotlight/Spotlight.tsx
```diff
@@ -275,32 +285,42 @@ const Spotlight = ({
...
+  }, [updateScrollPosition]);
 
   useEffect(() => {
     window.addEventListener('resize', updateWindowBounds);
 
     return () => {
       window.removeEventListener('resize', updateWindowBounds);
     };
-  }, []);
+  }, [updateWindowBounds]);
```
joien1, line 323: Could you use something like the top answer here to track the window size? 
https://stackoverflow.com/questions/62846043/react-js-useeffect-with-window-resize-event-listener-not-working

src/components/Spotlight/Spotlight.tsx
```diff
@@ -275,32 +285,42 @@ const Spotlight = ({
...
-    setScrollTop(e?.target?.documentElement.scrollTop);
-  };
+    // setScrollTop(e?.target?.documentElement.scrollTop);
+    setScrollTop(window.scrollY);
+  }, [setScrollTop]);
+
+  useEffect(() => {
+    updateScrollPosition(); // update scroll position every render that the scroll changes so the previous value will update after scrolling events stop
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [window.scrollY]);
```
joien1, line 305: I'm not confident that deps will react to changes on `window.xxxx` so I don't think this is doing what you expect.

src/components/Spotlight/Spotlight.tsx
```diff
@@ -275,32 +285,42 @@ const Spotlight = ({
...
 
   // TODO: use a resize observer to detect when the bounds of the target change
-  const updateWindowBounds = () => {
+  const updateWindowBounds = useCallback(() => {
     setWindowDimensions({
       width: window.innerWidth,
       height: window.innerHeight,
     });
-  };
+  }, [setWindowDimensions]);
```
joien1, line 293: Does this do anything? I thought that `setXX` state functions didn't get redefined across renders

src/components/Spotlight/Spotlight.tsx
```diff
@@ -275,32 +285,42 @@ const Spotlight = ({
...
 
-  const updateScrollPosition = (e: Event) => {
+  const updateScrollPosition = useCallback(() => {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore documentElement does exist on target element
-    setScrollTop(e?.target?.documentElement.scrollTop);
-  };
+    // setScrollTop(e?.target?.documentElement.scrollTop);
+    setScrollTop(window.scrollY);
```
joien1, line 299: https://www.npmjs.com/package/@react-hook/window-scroll

similar comment on using this for scroll position

joien1 (changes requested):
General comment is I think we can make custom hooks (or reuse ones others have made) for tracking things like scroll position and window size. Ignoring the `exhaustive-deps` rule is a big code smell for me and I think we can make things much simpler anyway with a dedicated hook for these attributes

src/components/Spotlight/Spotlight.tsx
aVileBroker: Yeah - I think I was just trying to fix what was already there, but a refactor is probably what is really needed.

aVileBroker:
Old Spotlight behavior:
https://user-images.githubusercontent.com/7891559/166563199-c5cd3729-ef07-44d9-92af-f15ed12bbafa.mp4

New Spotlight behavior:
https://user-images.githubusercontent.com/7891559/166563255-90e6ae4a-2cf2-4a45-818c-2ae156578eda.mp4

joien1 (approved):
Lgtm 🚢

PR #398, opened 2022-06-14
Remove nanoid dependency

dawsonbooth:
Replaces `nanoid()` call to generate an identifier with a call to a custom generator.

This removes the dependency on `crypto`, which causes issues for some CRA users:

- https://github.com/facebook/create-react-app/issues/11756
- https://github.com/webpack/webpack/issues/11282


src/components/InteractionFeedback/InteractionFeedback.tsx
```diff
@@ -1,13 +1,24 @@
...
+  const alphabet =
+    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+  let result = '';
+  for (let i = 0; i < length; i++) {
+    result += alphabet.charAt(Math.floor(Math.random() * alphabet.length));
+  }
+
+  return result;
+}
```
aVileBroker, line 21: Could you move this to a util file for reuse? Also arrow function plz 😋

aVileBroker (changes requested):

aVileBroker (approved):

PR #406, opened 2022-07-02
fix: webkit-appearance: none for ButtonContainer

### LINKED ISSUE #405: Button | Change -webkit-appearance to none to fix button anchors on Safari
For buttons with the `as` prop for the container changing it from a `button` to an `a`, keeps the role as `button` as expected, but safari forces button appearance to the anchor because of that role.

Fix is to set `-webkit-appearance: none;` on all button containers (we'll always want our styles)

Safari:
image

Chrome
image

asheerrizvi:
Closes #405

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -89,6 +89,7 @@ export const ButtonContainer: string & StyledComponentBase<any, {}, ButtonContai
     const fontColor = getFontColorFromVariant(variant, color, colors.background, colors.grayDark);
 
     return `
+      -webkit-appearance: none;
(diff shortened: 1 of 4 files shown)


aVileBroker:
@asheerrizvi you'll need to update your snapshots before pushing. You can run `yarn update-tests` to do that. Then commit the updated snapshot files.

asheerrizvi:
@aVileBroker done.

aVileBroker (approved):

aVileBroker:
Great, thanks so much for your contribution @asheerrizvi!

Issue #434, opened 2023-06-14
Table | Various fixes to support HS Constellation


meghanainaganti:
**Acceptance Criteria**
- Move responsive title out of cell
- Add styled responsive title (and other api)
- Ensure no title if name is undefined
- Responsive-view transitions don't work - only works on first render

xavierdonnellon:
Gameplan:
1. Move ResponsiveTitle to be sitting next to the CellComponent (so it is rendered even if the cellComponent is overridden).
2. Only show ResponsiveTitle if column name is not empty. 
3. add StyledResponsiveTitle parameter to Table, as well as responsiveTitleProps.

PR #436, opened 2023-06-16
Fixes #379 #434

### LINKED ISSUE #379: Types | StyledSubcomponentType is too rigid and complains about regular JSX
The intent behind the architecture is to allow any JSX to be passed into those `Styled_________` props. While we encourage restyling our internal base components with styled-components - we shouldn't complain if the developer doesn't do that, as flexibility is the selling-point of Foundry UI.

In the StyledSubcomponentType type, I would try to union the current type with `JSX.Element`.

xavierdonnellon:
**This PR accomplishes the following:** 
- Adds type support for custom component builders with props for columns in the `Columns` type, for things like cellComponent, headerCellComponent, etc, while retaining support for StyledComponents (see `ComponentBuilder`)
- Fixes group headers appearance in responsive view
- Makes custom sortFunctions functional in both directions, and extracts reusable comparison code into the `compareEntries` function, fixes group sorting
- Fixes sorting on the Table storyboard
- Fixes responsive view 
- Fixes passing in of rowProps StyledRow for non-grouped Table's
- Allows for styling of the ResponsiveHeaderCell, and wraps cell in CellContainer (which contains both the ResponsiveHeaderCell if in responsive mode, and the Cell).
- Change type of table `data` to `RowEntry[]`. Originally was `ColumnTypes`, which has no overlap with expected input. More clear typing.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -110,12 +110,11 @@ export const ResponsiveHeaderCell = styled(StyledBaseSpan)`
     const { colors } = useTheme();
     return `
       display: block;
-      padding: 1em 0;
       word-break: break-word;
       hyphens: auto;
       color: ${colors.primary};
-      padding: 0.5em;
       user-select: none;
+      padding: 0.5em;
       cursor: pointer;
       margin-right: .5em;
       background-color: rgba(0,0,0,0.05);
@@ -177,6 +176,7 @@ export const Cell = styled(StyledBaseTD)`
   word-break: break-word;
   hyphens: auto;
   width: unset;
+  padding: 0.5em;
 `;
 
 export const SortIcon = styled(Icon)`
@@ -192,7 +192,7 @@ export const SortIcon = styled(Icon)`
 
 const CellContainer = styled(StyledBaseDiv)`
   display: flex;
-  padding: 1em 0;
+  padding: 0.5em 0;
 `;
 
 /** Start of variables */
@@ -743,6 +743,7 @@ Table.Cell = Cell;
  */
 Table.Title = ResponsiveHeaderCell;
 Table.ResponsiveHeaderCell = ResponsiveHeaderCell;
+Table.CellContainer = CellContainer;
 Table.ExpansionIconColumnName = ExpansionIconColumnName;
 
 export default Table;
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -1,9 +1,10 @@
 import React, { useEffect, useState } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import useResizeObserver from 'use-resize-observer/polyfilled';
 import { mdiArrowDown, mdiChevronDown, mdiChevronRight, mdiChevronUp } from '@mdi/js';
 import Icon from '@mdi/react';
 import {
+  StyledBaseDiv,
   StyledBaseSpan,
   StyledBaseTable,
   StyledBaseTD,
@@ -12,7 +13,8 @@ import {
 } from '../../htmlElements';
 import {
   CellOptions,
-  columnTypes,
+  ColumnType,
+  ColumnTypes,
   InternalExpansionIconProps,
   RowProps,
   TableProps,
@@ -20,7 +22,7 @@ import {
 import { useAnalytics, useTheme } from '../../context';
 import { mergeRefs } from '../../utils/refs';
 
-type collapsedState = Record<string, string>;
+type CollapsedState = Record<string, string>;
 
 /** Start of styled components */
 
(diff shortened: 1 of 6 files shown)


src/components/Table/types.ts
```diff
@@ -11,30 +11,45 @@ export type InternalExpansionIconProps = {
...
+  footerContent?: ReactNode;
+  sortFunction?: (item1: any, item2: any) => boolean;
+  isGroupLabel?: boolean;
+  cellComponent?: ComponentBuilder;
+  rowComponent?: ComponentBuilder;
+  headerCellComponent?: ComponentBuilder;
+  groupCellComponent?: ComponentBuilder;
 }
 
+type ComponentBuilder = StyledSubcomponentType | ((props: any) => JSX.Element) | React.FC;
```
aVileBroker, line 32: Didn't we move this to a separately exported type?

src/components/Table/types.ts
```diff
@@ -11,30 +11,45 @@ export type InternalExpansionIconProps = {
...
+
+/**
+ * @deprecated Use `ColumnTypes`.
```
aVileBroker, line 35:
```suggestion
 * @deprecated Use `Columns`.
```

src/components/Table/types.ts
```diff
@@ -11,30 +11,45 @@ export type InternalExpansionIconProps = {
...
+/** The data for a single row. */
+export type RowEntry = Record<string, any>;
+
 export type TableProps = {
   areGroupsCollapsible?: boolean;
   columnGap?: string;
-  columns: columnTypes;
-  data?: columnTypes[] | Array<Array<columnTypes>>;
+  columns: Columns;
+  data?: RowEntry[] | Array<Array<RowEntry>>;
```
aVileBroker, line 46:
```suggestion
  data?: Array<RowEntry> | Array<Array<RowEntry>>;
```

or 

```suggestion
  data?: Rows | Array<Rows>;
```

src/components/Table/types.ts
```diff
@@ -43,34 +58,53 @@ export type TableProps = {
   StyledGroupLabelRow?: StyledSubcomponentType;
   StyledHeader?: StyledSubcomponentType;
   StyledHeaderCell?: StyledSubcomponentType;
+  /**
+   * If in responsive mode (width < `minWidthBreakpoint`), the `StyledResponsiveHeaderCell` will appear next to the
+   * `StyledCell`. By default, it displays the `name` of the column. This component is not displayed if not in responsive
+   * mode, or if the column has no `name`.
+   */
```
aVileBroker, line 65:
```suggestion
  /**
   * Header cells used when the Table is in condensed view
   * based on `minWidthBreakpoint`
   */
```

src/components/Table/types.ts
```diff
@@ -43,34 +58,53 @@ export type TableProps = {
...
+  StyledResponsiveHeaderCell?: StyledSubcomponentType;
   StyledRow?: StyledSubcomponentType;
   StyledFooter?: StyledSubcomponentType;
   StyledFooterCell?: StyledSubcomponentType;
+  /**
+   * If in responsive mode (width < `minWidthBreakpoint`), the `CellContainer` contains the `StyledCell`
+   * and the `StyledResponsiveHeaderCellA `CellContainer` contains the `StyledCell`, as well as the `StyledResponsiveHeaderCell` for each cell
+   * if width is less than `minWidthBreakpoint`. If width is greater than `minWidthBreakpoint`, then the `CellContainer`
+   * only wraps the `StyledCell`.
+   */
```
aVileBroker, line 75:
```suggestion
  /**
   * Extra wrapper around cells, which also contains the ResponsiveHeaderCell when in the condensed view
   */
```

src/components/Table/types.ts
```diff
@@ -43,34 +58,53 @@ export type TableProps = {
...
   columnGap: string;
   columnWidths: string;
   rowNum?: number;
   reachedMinWidth?: boolean;
   isCollapsed?: boolean;
 };
+
 export type CellOptions = {
-  RenderedCell: any;
+  RenderedCell: ComponentBuilder;
```
aVileBroker, line 104: Should be StyledSubComponent?

aVileBroker (changes requested):

aVileBroker (dismissed):

src/components/Table/Table.tsx
```diff
@@ -176,7 +176,7 @@ export const Cell = styled(StyledBaseTD)`
   word-break: break-word;
   hyphens: auto;
   width: unset;
-  padding: 0.5em;
+  padding: 0.5em 0;
```
aVileBroker, line 179: Why this change?

src/components/Table/types.ts
xavierdonnellon: Re-added to StyledSubcomponentType

src/components/Table/Table.tsx
xavierdonnellon: No need for horizontal padding on cells (makes it not line up with header)

xavierdonnellon: Also, half the padding comes from the Cell itself (the space between the text and the CellContainer) and half the padding went into the CellContainer (space between two rows)

aVileBroker (approved):

PR #443, opened 2023-06-27
Table sort improvements 2

xavierdonnellon:
This PR improves table sorting performance by not copying all table data each sort, but rather once on the first sort and re-using the copied reference.

It also fixes the isSorted prop passed into HeaderCell's, which was not always accurate.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -395,12 +395,27 @@ const Table = ({
     setCollapsedGroups(defaultCollapsed);
 
     if (newDirection === SortDirection.noSort) {
-      setSortedData(data); // reset data
+      setSortedData(data); // reset data to original
       return;
     }
 
-    // Shallow copy original, unsorted data so that `data` prop remains unmodified
-    const copiedData = [...data];
+    let copiedData: RowEntry[] = [];
+
+    // Make shallow copy if references are equal
+    if (sortedData === data) {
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.tsx
```diff
@@ -395,12 +395,27 @@ const Table = ({
...
       return;
     }
 
-    // Shallow copy original, unsorted data so that `data` prop remains unmodified
-    const copiedData = [...data];
+    let copiedData: RowEntry[] = [];
+
+    // Make shallow copy if references are equal
+    if (sortedData === data) {
+      console.log('Copying data for sorting');
```
aVileBroker, line 406:
```suggestion
```

src/components/Table/Table.tsx
```diff
@@ -395,12 +395,27 @@ const Table = ({
...
 
     if (newDirection === SortDirection.noSort) {
-      setSortedData(data); // reset data
+      setSortedData(data); // reset data to original
       return;
     }
 
-    // Shallow copy original, unsorted data so that `data` prop remains unmodified
-    const copiedData = [...data];
+    let copiedData: RowEntry[] = [];
```
aVileBroker, line 402: maybe `localData` is more applicable?

aVileBroker (changes requested):

aVileBroker (approved):

Issue #437, opened 2023-06-21
Avatar Component


aVileBroker:
As a user of FoundryUI, I want a component that is able to show an image in a circle/squircle/square/etc. It should show initials in place of my profile photo if it isn't passed in.

### AC:

* New Avatar component
* Includes states for
  * has image url
  * loading
  * error
  * no image url
* Has a variable size
* User Initials, padding, border, etc. scale with the size of the component

>Consistent look/feel/interaction across browsers
>0 A11y violations
>90%+ code coverage
>Storybook knobs, actions, and docs where appropriate
>Responsive

shuvrojit:
Assign it to me.

PR #444, opened 2023-07-03
Avatar Component (Issue #437)

meghanainaganti:
-Creates an avatar component

[jira ticket](https://headstorm.atlassian.net/browse/CON-1356?atlOrigin=eyJpIjoiNGM0ZTU1YmUzOGVmNDJhMTgxMjI5NTAyNDljYTNkYWUiLCJwIjoiaiJ9)

[github issue](https://github.com/Headstorm/foundry-ui/issues/437#issue-1768405724)


aVileBroker:
You'll need to add the component to the index.ts of the repo, you can also test it out in the code sandbox in the other comment

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+import Skeleton from '../Skeleton/Skeleton';
+
+export type AvatarProps = {
+  initials?: ReactNode;
+  imgURL?: string;
+  size?: number;
+  shape?: number;
+  isLoading?: boolean;
+  isError?: boolean;
+  hasImage?: boolean;
```
aVileBroker, line 15: Can we not derrive this based on image Url?

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  StyledAvatarContainer?: StyledSubcomponentType;
+  StyledAvatarText?: StyledSubcomponentType;
+  StyledAvatarImage?: StyledSubcomponentType;
+
+  avatarProps?: SubcomponentPropsType;
+
+  avatarTextRef?: React.RefObject<HTMLSpanElement>;
```
aVileBroker, line 23: Nit: keep the same order as the other props (container, text, image)

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+  shape?: number;
+  isLoading?: boolean;
+  isError?: boolean;
+  hasImage?: boolean;
+
+  StyledAvatarContainer?: StyledSubcomponentType;
+  StyledAvatarText?: StyledSubcomponentType;
+  StyledAvatarImage?: StyledSubcomponentType;
+
+  avatarProps?: SubcomponentPropsType;
```
aVileBroker, line 21: There should be a "props" prop for each subcomponent

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+  avatarImageRef?: React.RefObject<HTMLImageElement>;
+  avatarContainerRef?: React.RefObject<HTMLDivElement>;
+};
+
+export const AvatarContainer = styled(StyledBaseDiv)`
+  ${({ size, shape }: { size: number; shape: string }) => {
+    return `
+      display: flex;
+      border-radius: ${`${shape}%`};
+      background-color: #eaeaea;
```
aVileBroker, line 33: Color should come from useTheme

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+      width: ${`${size * 3}rem`};
+      height: ${`${size * 3}rem`};
+      justify-content: center;
+      align-items: center;
+    `;
+  }}
+`;
+
+export const AvatarText = styled(StyledBaseSpan)`
+  ${({ size, colors }: { size: string; colors: string }) => {
```
aVileBroker, line 44: Colors makes it sound like there are multiple, just use the singular or specify "text Color"

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+export const AvatarContainer = styled(StyledBaseDiv)`
+  ${({ size, shape }: { size: number; shape: string }) => {
+    return `
+      display: flex;
+      border-radius: ${`${shape}%`};
+      background-color: #eaeaea; 
+      padding: 1rem;
+      width: ${`${size * 3}rem`};
```
aVileBroker, line 35: Instead of needing to calculate relative sizes, use "em" - the text can be .33em and the container can be 3em (and set the font size to 1em)

src/components/Avatar/_tests_/Avatar.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+const testId = 'foundry-test';
+
+describe('Avatar', () => {
+  it('matches snapshot', async () => {
+    // const { container, getByTestId } = render(<Avatar />);
+    // await waitFor(() => getByTestId(testId));
+    // expect(container).toMatchSnapshot();
+  });
+
+  it('', async () => {});
```
aVileBroker, line 17: Probably needs some tests here, or not

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+  avatarProps = {},
+  avatarContainerRef,
+  avatarTextRef,
+  avatarImageRef,
+}: AvatarProps): JSX.Element => {
+  const { colors } = useTheme();
+  if (isLoading) {
+    return (
+      <div style={{ borderRadius: `${shape}%`, overflow: 'hidden' }}>
+        <Skeleton isLoading color={colors.grayLight}>
```
aVileBroker, line 93: Instead of a div wrapping the skeleton wrapping another div, you can make the container a style skeleton shimmer (see the skeleton story for an example)

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <div style={{ width: `${size * 3}rem`, height: `${size * 3}rem` }} />
+        </Skeleton>
+      </div>
+    );
+  }
+
+  if (hasImage) {
+    return (
+      <StyledAvatarImage ref={avatarImageRef} size={size} shape={shape}>
+        <img alt="profile" {...avatarProps} src={imgURL} ref={avatarImageRef} />
```
aVileBroker, line 103: This can just be a background image on the container to further simplify the html

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </StyledAvatarImage>
+    );
+  }
+
+  return (
+    <StyledAvatarContainer ref={avatarContainerRef} size={size} shape={shape}>
```
aVileBroker, line 109: Given shape only changes the border radius in percent, I would change the name to reflect that: "borderRadiusPercent" or something similar.

aVileBroker (changes requested):

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  ${({ borderRadiusPercent }: Pick<AvatarProps, 'borderRadiusPercent'>) =>
+    `
+    border-radius: ${borderRadiusPercent}%;
+  `}
+`;
+
+const Avatar = ({
+  placeholder,
+  // eslint-disable-next-line no-unused-vars
+  children,
```
aVileBroker, line 90: looks like you're not passing children through - you'll need to add it to the jsx around line 125

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  imgURL,
+  size = 3,
+  borderRadiusPercent,
+  color = colors.grayXlight,
```
aVileBroker, line 94: Best not to default the color here, as it should come from the useTheme hook

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,130 @@
...
+  isLoading,
+  StyledAvatarContainer = AvatarContainer,
+  StyledAvatarText = AvatarText,
+  StyledAvatarShimmer = AvatarShimmer,
+  avatarTextProps = {},
+  avatarContainerRef,
+  avatarTextRef,
```
aVileBroker, line 101: you're missing some of the props you specified in the proptypes, the container props should be spread over the skeleton containerProps

aVileBroker (changes requested):

src/components/Avatar/__tests__/Avatar.test.tsx
```diff
@@ -2,27 +2,28 @@ import React from 'react';
...
+import colors from '../../../enums/colors';
 
 expect.extend(toHaveNoViolations);
 configure({ testIdAttribute: 'data-test-id' });
 const testId = 'foundry-avatar';
 
 describe('Avatar', () => {
   it('matches snapshot', async () => {
-    const { container } = render(<Avatar />);
+    const { container } = render(<Avatar color={colors.grayXlight} />);
```
aVileBroker, line 13: You can still default color to `colors.grayXlight`, just do it after the `useTheme` hook call

aVileBroker (changes requested):

aVileBroker:
Playing with it in the codesandbox, I think the default size is wayyyy too big. I would expect it to be 1 or 1.5. Here I added it to the h1.

I also wonder if we should make it inline-block by default... I'm really not sure either way, it would just make my specific example where I didn't use a flexbox better so it may not be necessary.
image
image

src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <LabelContainer>{children}</LabelContainer>
+  </Avatar>
+);
+
+LabelAvatar.args = {
+  placeholder: 'JS',
+  children: 'John Smith',
+  size: 10,
+  imgURL: 'https://source.unsplash.com/collection/19271953',
+  borderRadiusPercent: 0,
```
aVileBroker, line 43: maybe like 10% or something

aVileBroker (dismissed):

aVileBroker (approved):

PR #456, opened 2023-07-17
451 dropdown value count

j-leidy:
This branch is also up to date with 450-dropdown-search-improvements. 

- Prop added to show value count
- Created ValueCount subcomponent
- Made valueCount style-able, ref-able, prop-able
- Added variant support
- Renamed closeIcons to infoIcons


src/components/Dropdown/Dropdown.tsx
```diff
@@ -231,6 +232,17 @@ const SearchInput = styled(StyledBaseInput)`
   text-align: left;
 `;
 
+const ValuesCountContainer = styled(StyledBaseDiv)`
+  ${({ variant, color, dropdownVariant }: UsefulDropdownState & { dropdownVariant: variants }) => {
+    return `
+    ${getDropdownTagStyle(dropdownVariant, variant, color, '')}
+      padding: 0.125rem;
+      border-radius: 1.5rem;
+
```
aVileBroker, line 241:
```suggestion
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -786,9 +810,47 @@ const Dropdown = ({
     [rememberScrollPosition],
   );
 
-  const closeIcons = (
-    <>
-      {handleClear && values.length > 0 && (
+  const valueCountCloseIconHandler = () => {
+    if (showValueCount && showCloseIcon) {
```
aVileBroker, line 814: shouldn't these be split so showValueCount only shows/hides the value count? Otherwise why have two flags if both need to be true to see either of them

src/components/Dropdown/Dropdown.tsx
```diff
@@ -786,9 +810,47 @@ const Dropdown = ({
...
+            tabIndex={tabIndex}
+            ref={closeIconRef}
+            {...closeIconProps}
+          >
+            <Icon path={mdiClose} size="1em" />
+          </StyledCloseIconContainer>
+        </>
+      );
+    }
+    if (showValueCount && !showCloseIcon) {
```
aVileBroker, line 839: I think you're way overcomplicating this lol

aVileBroker (changes requested):

src/components/Dropdown/Dropdown.tsx
j-leidy: Yeah the if block seemed like there had to be a better way to me.. Looking into it.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -294,13 +313,16 @@ export interface DropdownProps {
...
   valueVariant?: variants;
+  valueCountVariant?: variants;
 
   shouldStayInView?: boolean;
   intersectionThreshold?: number;
   intersectionContainer?: HTMLElement | null;
   intersectionObserverPrecision?: number;
   virtualizeOptions?: boolean;
 
+  showCloseIcon?: boolean;
```
aVileBroker, line 324: this should be covered by the clearable prop - there shouldn't be a use-case where the dropdown is clearable but the clear button isn't shown (making it not clearable)

aVileBroker:
Playing with the [sandbox](https://codesandbox.io/s/react-forked-cyv25m?file=/src/App.js), I think `showValuesCount` should default to false

src/components/Dropdown/Dropdown.tsx
j-leidy: Good call I agree. Changing now

j-leidy: It looks like clearable is appended onto DropdownProps in the story. Removing from there and adding to DropdownProps.

aVileBroker: cool

aVileBroker (approved):

Issue #441, opened 2023-06-23
RangeSlider | Make onDrag more in-line with other onChange handlers like TextInput


CameronMillerSMU:
### AC
- Deprecate onDrag and change it to be onDebounceChange
- onDrag will point to onDebounceChange
- add onChange which isn't limited by debounce
- Fix problem with drag handle not being centered vertically

PR #459, opened 2023-07-18
Improve range slider callbacks

xavierdonnellon:
In response to issue #441

This PR adds to RangeSlider:
 - onChange callback, called immediately on every value change to slider
 - onDebounceChange callback, called after debounceInterval 
 - Deprecates onDrag, but maintains its functionality

Other changes to RangeSlider:
 - Change DragHandle's beingDragged to $beingDragged (transient StyledComponent prop to suppress DOM warning that beingDragged is an unknown attribute) 
 - Improves typing of processedMarkers and processedValues to remove type errors and redundant code

### DIFF
--- a/src/components/RangeSlider/RangeSlider.tsx
+++ b/src/components/RangeSlider/RangeSlider.tsx
@@ -67,7 +67,7 @@ export const Container = styled.div`
 `;
 
 export const DragHandle = styled(a.div)`
-  ${({ beingDragged = false, color }: HandleProps) => {
+  ${({ $beingDragged = false, color }: HandleProps) => {
     const { colors } = useTheme();
     const handleColor = color || colors.primary;
     return `
@@ -87,7 +87,7 @@ export const DragHandle = styled(a.div)`
 
       filter: url(#blur);
 
-      cursor: ${beingDragged ? 'grabbing' : 'grab'};
+      cursor: ${$beingDragged ? 'grabbing' : 'grab'};
       z-index: 2;
     `;
   }}
@@ -254,7 +254,7 @@ export const RangeSlider = ({
     }
   }
 
-  const { colors } = useTheme();
+  // const { colors } = useTheme();
   const { prefersReducedMotion } = useAccessibilityPreferences();
   const hasHandleLabels = useRef(false);
   const initializing = useRef(true);
@@ -273,21 +273,21 @@ export const RangeSlider = ({
     () =>
       values?.map((val: number | ValueProp) => {
         if (typeof val === 'number') {
-          return { value: val, label: val, color: colors.primary };
+          return { value: val, label: undefined, color: undefined };
         }
         if (Object.prototype.hasOwnProperty.call(val, 'label')) {
           hasHandleLabels.current = true;
         }
         return val;
       }) ?? [],
-    [values, colors],
+    [values],
   );
 
   // Convert `markers` prop from `number[] | ValueProp[]` into strictly `ValueProp[]`
   const processedMarkers: Array<ValueProp> =
     markers?.map((val: number | ValueProp) => {
       if (typeof val === 'number') {
-        return { value: val, label: val, color: colors.primary };
+        return { value: val, label: undefined, color: undefined };
       }
       return val;
     }) ?? [];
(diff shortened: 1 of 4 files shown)


src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -366,48 +348,53 @@ export const RangeSlider = ({
...
-      debouncedDrag();
-      set({
-        x: down ? deltaX : pixelPositions[0],
-        y: down ? deltaY : 0,
+      handleDrag(valueBuffer.current);
 
-        immediate: prefersReducedMotion || springOnRelease ? down : true,
+      let animate = true;
+      if (prefersReducedMotion) animate = false;
+      if (!snapToValue) animate = springOnRelease ? !down : false;
```
aVileBroker, line 389: Trying to simplify your logic here, probably not right but you get the idea
```suggestion
      let animate = !down;
      animate = prefersReducedMotion || !springOnRelease || snapToValue ? false : animate;
```

aVileBroker (changes requested):
Tons of changes to review, but wanted to give you something to start on

stats/statistics.html
aVileBroker, line 1: I thought we were gitignoring this? Or do the diffs look pretty small when they happen?

src/components/RangeSlider/__tests__/RangeSlider.test.tsx
```diff
@@ -38,7 +39,7 @@ describe('RangeSlider', () => {
 
   describe('Accessibility Tests', () => {
     it('Should pass accessibility test with default props', async () => {
-      const component = <RangeSlider min={0} max={10} testId={testId} />;
+      const component = <RangeSlider min={0} max={10} testId={testId} values={[1]} />;
```
aVileBroker, line 42: this should probably be a separate test outside of the "default props" test

aVileBroker (changes requested):

src/components/RangeSlider/types.ts
```diff
@@ -64,16 +65,34 @@ export type RangeSliderProps = {
...
   max: number;
-  values?: number[] | ValueProp[];
+  values: number[] | ValueProp[];
   testId?: string;
   markers?: number[] | ValueProp[];
+  /**
+   * If true, the drag handle will snap to the passed-in `value`.
+   * If false, the drag handle will directly follow the mouse on drag.
+   */
+  dragHandleBehavior?: 'snapToValue' | 'followMouse';
```
aVileBroker, line 90: I feel like camelCase isn't right for text attributes - but maybe we could reduce it to one word either way, like
```suggestion
  dragHandleAttachment?: 'value' | 'mouse';
```

src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -226,118 +215,109 @@ export const RangeSlider = ({
...
+    debounce(newVal => {
+      if (onDrag) onDrag(newVal);
+      if (onDebounceChange) onDebounceChange(newVal);
+    }, debounceInterval),
+  ).current;
+
+  const debouncedOnRelease = useRef(
+    // wait an extra ms. onRelease should be called after onChange
+    debounce(newVal => onRelease && onRelease(newVal), debounceInterval + 1),
+  ).current;
```
aVileBroker, line 257: What's up with these refs? Like an alternative to useCallback?

aVileBroker (changes requested):

stats/statistics.html
xavierdonnellon: Accidentally gitignored wrong thing

src/components/RangeSlider/RangeSlider.tsx
xavierdonnellon: They're not just callbacks, they actually are callable objects (they have other funcs on them like debouncedOnChange.cancel() and stuff, which I just added disposing)

src/components/RangeSlider/RangeSlider.tsx
xavierdonnellon: There is some point in the truth table which this doesn't match up. I think to accomplish above we'd need nested ternaries which causes a warning and prob isn't worth it

PR #466, opened 2023-07-21
Fix | Dropdown ArrowIconContainer isOpen Prop Structure

j-leidy:
- Modified the isOpen prop structure for the ArrowIconContainer. 
- Allows for cleaner access to the prop during implementation.


aVileBroker (commented):
Could you make this a transient prop btw? @xavierdonnellon can give some direction on that

aVileBroker (approved):

PR #469, opened 2023-07-28
Implemented isReadOnly prop for RangeSlider component

ishaanbhagwat:
- This PR implements isReadOnly prop to the RangeSlider.
-This prop ensures that the RangeSlider cannot be used to change the value in readOnly mode.
- This feature can be really useful when implementing the RangeSlider and ensuring that only certain users can edit the value on the RangeSlider.

### DIFF
--- a/src/components/RangeSlider/RangeSlider.stories.tsx
+++ b/src/components/RangeSlider/RangeSlider.stories.tsx
@@ -74,7 +74,7 @@ export const Default: Story<DefaultProps> = ({
   max,
   debounceInterval,
   axisLock,
-  maintainColorWhenDisabled,
+  isReadOnly,
 }: DefaultProps) => {
   const [val, setVal] = useState(value);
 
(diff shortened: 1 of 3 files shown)


src/components/RangeSlider/RangeSlider.stories.tsx
```diff
@@ -74,6 +74,7 @@ export const Default: Story<DefaultProps> = ({
   max,
   debounceInterval,
   axisLock,
+  isReadOnly,
```
aVileBroker, line 77: should be `readonly` to match the pattern set by HTML inputs

src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -40,9 +46,15 @@ export const Container = styled.div`
...
-    `
+        filter: grayscale(1) contrast(.5) brightness(1.2);
+        pointer-events: none;`
+        : ''
+    }
+
+    ${
+      isReadOnly
+        ? `
+        pointer-events: none;`
```
aVileBroker, line 57: we want to keep pointer events, so things like "hover to see handle label" still work, but other mouse/kb inputs don't change the values.

aVileBroker (changes requested):

aVileBroker (approved):

aVileBroker (approved):

PR #477, opened 2023-10-06
Fixes #476 Smooth scrolling to new target's location

### LINKED ISSUE #476: Spotlight | Scroll to element if not visible
AC
- [ ] Use the Intersection Observer API to figure out if its off screen
- [ ] figure out the scroll position
- [ ] scroll to position
- [ ] use custom element if the body isn't what gets scrolled in the app (or walk up the DOM to find the scrollable container??)
- [ ] include a scroll padding attribute to set how much extra space to give the element on top of its scroll position so it doesn't literally touch it

aVileBroker:


daniel-grimm (approved):

src/components/Spotlight/Spotlight.tsx
```diff
@@ -237,10 +289,10 @@ const Spotlight = ({
       lightPath: finalRectangularPath,
       circularLightPath: circularPath,
 
-      annotationTransform: `translate(${rect.x}px, ${rect.y}px) translate(0%, -100%)`,
+      annotationTransform: `translate(${rect.x}px, ${rect.y}px) translate(0%, -100%)`, // TODO: change second translate to make different attach positions
```
GarrettRathke, line 292: @aVileBroker This looks great! Thanks for doing this :-)
I have a few questions about this comment. Is this something that's too big to include in this PR? Or perhaps does it just make more sense in another PR? And if so, is there another Jira ticket for this work? Cheers!

aVileBroker: This would be the other ticket, I can reference it in the comments
https://github.com/Headstorm/foundry-ui/issues/315

GarrettRathke (approved):
Thank you! Good job!

PR #478, opened 2023-10-12
Fixes #476 Smooth scrolling to new target's location

### LINKED ISSUE #476: Spotlight | Scroll to element if not visible

GarrettRathke:
This is based off of @aVileBroker's work from https://github.com/Headstorm/foundry-ui/pull/477. I just reverted some dependency changes and cleaned up some minor type errors in my fork.


aVileBroker (approved):
squeaky clean, thanks!

